PR #130, opened 2014-11-06
Gracefully handle VPCs with no Name tag

jwm:
I missed one instance where we don't gracefully handle VPCs with no Name tag.


riccardomurri:
As far as I can understand, current code implements the same feature differently.

PR #132, opened 2014-11-06
Calculate AWS key fingerprints the same way AWS does

jwm:
Thanks to these forum posts:

https://forums.aws.amazon.com/message.jspa?messageID=307410
http://stackoverflow.com/questions/21709817/printing-pem-key-der-form-in-python

I added support for calculating AWS key fingerprints, which eliminates the warning on cluster start.


riccardomurri:
Superceded by #536 where the code is reworked to align with current "master".  Conversation continues there if you're still interested.

Many many thanks for taking the time to package this contribution as a PR, and please accept my apologies for taking soooo much time to review and merge it!

PR #133, opened 2014-11-16
EC2: support customization of mounted root volume (size, storage type, iops)

jwm:
It's nice to be able to customize the root volume size and storage type. This branch does that, using a patch submitted by @chapmanb.


riccardomurri:
Superceded by #536 where the code is reworked to align with current "master".  Conversation continues there if you're still interested.

Many many thanks for taking the time to package this contribution as a PR, and please accept my apologies for taking soooo much time to review and merge it!

PR #143, opened 2015-01-16
Support Amazon AMIs: identify as CentOS and required tweaks

jwm:
These are the changes we identified so far to work with [Amazon Linux](http://aws.amazon.com/amazon-linux-ami/). The cluster doesn't start (the instances power on, stay `running` for a couple seconds, then shut down with an empty console log) but these changes will be necessary.

We don't have any plans to pursue this right now, but thought I'd open a PR in case anyone wants to pick this up.


riccardomurri:
It would be relatively easy to merge these changes into "master" but is Amazon Linux still a thing?  Why would one want to use it instead of the "real" CentOS?

PR #144, opened 2015-01-16
Fetch hostkeys securely from launched instances

jwm:
Hi Antonio, this implements the changes you suggested in https://github.com/gc3-uzh-ch/elasticluster/issues/116 to fetch SSH host keys from EC2 instances via the AWS API.

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -580,6 +580,13 @@ Optional configuration keys
     The maximum number of process to be created when virtual machines
     are started. Default is 10.
 
+``ssh_hostkeys_from_console_output``
+
+    whether to retrieve SSH host keys from instances' console output,
+    instead of accepting the host keys without verification when connecting
+    via SSH for the first time. Only supported when the cloud provider is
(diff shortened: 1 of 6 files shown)


arcimboldo:
Sorry for the late reply, we are busy building our new private cloud and have little time for Elasticluster in these months...

There are a few issues with this pull request:
* it only supports ec2_boto cloud provider, while also openstack and gce have similar call to get the console.
* it doesn't merge cleanly (well, that's probably our fault, because we didn't have time to check it until now), please re-submit after rebasing on the current master branch
* option `ssh_hostkeys_from_console_output` should be part of the signature of Cluster.__init__, instead of reading its value from the `extra` dictionary.

pinguinkiste:
I made a PR with an alternative solution that is much simpler. Feedback welcome.

Best,

Manuel

jwm:
I (obviously) haven't had a chance to move forward with this PR, and #269 looks like a better evolution of this code.

PR #157, opened 2015-05-01
Update the instructions for obtaining the client_id and client_secret for Google Compute Engine

mbookman:
Also removed the note in the Google-specific section regarding image_user as it is called out in the generic "Mandatory configuration keys".


mbookman:
Re-submitted PR as #164 to isolate the doc changes.
#157 inadvertently pulled in #158.

PR #158, opened 2015-05-04
Adds support for "boot_disk_type" and "boot_disk_size" instance configuration parameters.

mbookman:
Support is added for Google Compute Engine (gce.py).
Should be extendable to other providers.

### DIFF
--- a/elasticluster/providers/gce.py
+++ b/elasticluster/providers/gce.py
@@ -212,7 +212,10 @@ class GoogleCloudProvider(AbstractCloudProvider):
                        username=None,
                        # these params are specific to the
                        # GoogleCloudProvider
-                       instance_name=None, **kwargs):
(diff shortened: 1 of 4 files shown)


nicolasbaer:
It looks like this change is already included in your other pull request #157

elasticluster/providers/gce.py
```diff
@@ -212,7 +212,10 @@ def start_instance(self,
                        username=None,
                        # these params are specific to the
                        # GoogleCloudProvider
-                       instance_name=None, **kwargs):
+                       instance_name=None,
+                       boot_disk_type='pd-standard',
```
nicolasbaer: docstring is missing for the new parameters, but I will look into this when I merge in the changes from #133.

PR #166, opened 2015-06-08
Adds support for short-hand references to GCE images

mbookman:
such as:
  backports-debian
  ubuntu
  container-vm
These image names do not readily map to their associated Cloud project
name in the same way that debian, centos, and others do.


nicolasbaer:
thanks @mbookman

PR #167, opened 2015-06-30
Add support for nonstandard ssh ports

dsteinkraus:
This is PR #1 of 2 for adding Microsoft Azure support to elasticluster.

The purpose of this PR is to add support for nonstandard ssh ports to elasticluster (as a generic
feature, independent of Azure). 

The second PR, coming shortly, will be to add the actual Azure cloud provider.

Why this first PR is needed: when Azure creates a cluster, all the virtual machines 
have the same public IP, which is the address of a load balancer. Each of the VMs will 
have its own ssh port on the load balancer, which the load balancer in turn maps to port 
22 on the VM. So we need to allow each VM to return an IP:port pair which elasticluster 
will use to contact that VM.

Changes: if the get_ips() method of the cloud provider returns an address of the 
form address:port, elasticluster will: 

1. recognize that a port number has been provided;

2. use that port when it connects to the VM using Paramiko; and

3. in the Ansible inventory file, omit the port number from the host address, but 
specify it with the ansible_ssh_port variable.

Both IPv4 and IPv6 representations of addresses, with or without port numbers, are handled correctly.

In cases where the cloud provider doesn't return a port number (i.e. any cloud provider to date
other than Azure), both elasticluster and Ansible should work exactly as before.

This will prepare elasticluster to work with a cloud provider that uses nonstandard 
ssh ports and which may host multiple VMs on the same public IP address.

(Note: Azure does support something called "Public IPs", by which a VM can be exposed 
at its own internet IP address with ssh on port 22. However, this feature is presently 
only enabled for a maximum of 5 VMs per subscription, so it isn't very useful.)

The Azure cloud provider itself is completed and tested, and a second PR for that will follow soon.

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -38,6 +38,9 @@ from elasticluster.exceptions import TimeoutError, NodeNotFound, \
     InstanceError, ClusterError
 from elasticluster.repository import MemRepository
 
+SSH_PORT = 22
+IPV6_RE = re.compile('\[([a-f:A-F0-9]*[%[0-z]+]?)\](?::(\d+))?')
+
 class IgnorePolicy(paramiko.MissingHostKeyPolicy):
     def missing_host_key(self, client, hostname, key):
         log.info('Ignoring unknown %s host key for %s: %s' %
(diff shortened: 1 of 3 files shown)


riccardomurri:
Superceded by PR #168

Issue #169, opened 2015-07-15
"resize" command fails on Google Cloud: "extra" params not passed to cluster.add_node


mbookman:
On Google Compute Engine I found that when issuing a "resize" command, the new nodes were created, but then failed to get setup and were removed.

Turning on verbose debugging found:

    DEBUG:gc3.elasticluster:Ignoring error (u'104.197.46.24', <paramiko.rsakey.RSAKey
    object at 0x7f767c7981d0>, <paramiko.rsakey.RSAKey object at 0x7f767c925b10>)
    connecting to compute002

Also found that the boot disk was not getting set to the custom size I had set (and was only being set to the default 10GB).

The root cause of both issues was that the "extra" params were not being passed in to cluster.add_node.

I have a PR for this, which I will submit momentarily...

PR #170, opened 2015-07-15
For "resize" command, pass the "extra" fields to cluster.add_node

mbookman:
I fixed this for both the case where the first node is used as a template as well as the case where an explicit template is specified. I moved two lines out of the "for" loop:

    image_user = conf['login']['image_user']
    userdata = conf_kind.get('image_userdata', '')

as best as I can tell, they don't need to be set with each iteration.

This is intended to fix https://github.com/gc3-uzh-ch/elasticluster/issues/169.


nicolasbaer:
Thank you @mbookman and yes, of course the two lines can live outside of the loop just fine :)

Issue #174, opened 2015-07-21
Install fails: error: pbr 0.11.0 is installed but pbr<2.0,>=1.3 is required by set(['oslo.utils'])


mbookman:
Running an install of Elasticluster in a clean local workspace currently fails with:

$ python setup.py install
...
Installed /usr/local/google/home/mbookman/scratch/workspace/elasticluster.clean/lib/python2.7/site-packages/monotonic-0.2-py2.7.egg
error: pbr 0.11.0 is installed but pbr<2.0,>=1.3 is required by set(['oslo.utils'])

I'll submit a PR for this shortly. Simple fix is to update setup.py:

-    'pbr==0.11.0',
+    'pbr>=0.11.0',

but we should be able to just remove the version completely here now. After changing the above version 1.3 of pbr was installed and everything seems to work fine:

$ pip show pbr | grep ^Version
Version: 1.3.0

nicolasbaer:
relates to #162

nicolasbaer:
fixed in f055e2b

PR #175, opened 2015-07-21
Remove explicit pbr version number (0.11.0).

mbookman:
The explicit version causes install errors due to recent changes
to the oslo package:

Install fails: error: pbr 0.11.0 is installed but pbr<2.0,>=1.3 is required by set(['oslo.utils'])

This fixes https://github.com/gc3-uzh-ch/elasticluster/issues/174

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ required_packages = [
     'boto',
     # OpenStack clouds
     'python-novaclient',
-    'pbr==0.11.0',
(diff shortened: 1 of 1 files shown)


nicolasbaer:
We can remove the pbr dependency altogether. I only added it because there was a bug within one of the novaclient packages. It's fixed in f055e2b and therefore I close this request.

Issue #176, opened 2015-07-21
Removing nodes should remove records from known_hosts file


mbookman:
Have found that when I have a cluster in which I remove and add nodes that the adding of nodes often fails because Elasticluster cannot connect to the new node. The underlying problem is that the IP address for the new node is a re-use of a previous IP address. Thus the associated host key in the cluster's known_hosts file is the old one.

I'm not sure where the fix for this should go. It seems like it could be in node.stop() or cluster.remove_node(). It also looks to me like in order to remove a line from the cluster's known_hosts file requires using paramiko to recreate the file (I see no remove/delete method).

I do have a PR which I will send in a moment, which simply improves the messaging reported. This improved messaging better surfaces the underlying connection problem.

nicolasbaer:
That is something we did not handle yet, true. The logic should go into `cluster.remove_node` and yes, it looks like we would have to recreate the file.

riccardomurri:
Fixed by @whrc in #554

PR #177, opened 2015-07-21
Provides user-facing message when connect fails for host key error

mbookman:
This is important for clusters in which nodes are removed and added.
IP addresses can be reused and this failure to re-connect is currently hidden from users.

The *right* thing is probably to change "node.stop()" and/or "cluster.remove_node()"
functionality to remove entries from the known hosts file, but better error reporting here is a first step.

See: https://github.com/gc3-uzh-ch/elasticluster/issues/176

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -394,7 +394,7 @@ class Cluster(Struct):
         else:
             try:
                 node.start()
-                log.info("_start_node: node has been started")
+                log.info("_start_node: node '%s' has been started" % node.name)
                 return True
             except Exception as e:
                 log.error("could not start node `%s` for reason "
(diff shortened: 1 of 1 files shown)


elasticluster/cluster.py
```diff
@@ -954,6 +954,9 @@ def connect(self, keyfile=None):
             except socket.error as ex:
                 log.debug("Host %s (%s) not reachable: %s.",
                           self.name, ip, ex)
+            except paramiko.BadHostKeyException as ex:
+                print "Invalid host key: host %s (%s); check keyfile: %s" % (
```
nicolasbaer: how about we change this from `print` to `log.error()`?

mbookman: Yep. Done.

nicolasbaer:
Thanks @mbookman

PR #190, opened 2015-11-20
Modify gce.py get_ips() to return only the public IP address.

mbookman:
Elasticluster should never try to connect to the private IP address.
This can lead to problems as Elasticluster can connect to local machines
in the same IP range (and try to configure them).


riccardomurri:
We use private IP ranges in our on-campus OpenStack cloud, but I agree with you that they should never be used with public clouds.

Merged into the  `master` branch; thank you!

PR #208, opened 2015-12-01
Add support for Compute Engine instance tags

mbookman:
Add support for Compute Engine instance tags
 (https://cloud.google.com/compute/docs/reference/latest/instances#tags)


riccardomurri:
Manually merged.

Thanks!

PR #209, opened 2015-12-02
Support GCE networks

squee1945:
The code already supports GCE networks, but the configuration validation is preventing it from being used. This P-R simply adjusts the configuration validation to allow it.


riccardomurri:
Thanks, merged!

Issue #216, opened 2015-12-07
ImportError: No module named netifaces


mbookman:
Did a clean install of Elasticluster from source:

$ virtualenv elasticluster
$ source elasticluster/bin/activate
$ cd elasticluster
$ git clone git://github.com/gc3-uzh-ch/elasticluster.git src
$ cd src
$ python setup.py install

$ elasticluster list
Traceback (most recent call last):
  File "/usr/local/google/home/mbookman/scratch/workspace/elasticluster.gg_merge/bin/elasticluster", line 9, in <module>
    load_entry_point('elasticluster==1.3.dev0', 'console_scripts', 'elasticluster')()
  File "/usr/local/google/home/mbookman/scratch/workspace/elasticluster.gg_merge/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 546, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/google/home/mbookman/scratch/workspace/elasticluster.gg_merge/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2666, in load_entry_point
    return ep.load()
  File "/usr/local/google/home/mbookman/scratch/workspace/elasticluster.gg_merge/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2339, in load
    return self.resolve()
  File "/usr/local/google/home/mbookman/scratch/workspace/elasticluster.gg_merge/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2345, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "build/bdist.linux-x86_64/egg/elasticluster/__init__.py", line 33, in <module>
  File "build/bdist.linux-x86_64/egg/elasticluster/providers/openstack.py", line 31, in <module>
  File "/usr/local/google/home/mbookman/scratch/workspace/elasticluster.gg_merge/local/lib/python2.7/site-packages/python_novaclient-2.35.0-py2.7.egg/novaclient/client.py", line 38, in <module>
    from oslo_utils import netutils
  File "/usr/local/google/home/mbookman/scratch/workspace/elasticluster.gg_merge/local/lib/python2.7/site-packages/oslo.utils-3.1.0-py2.7.egg/oslo_utils/netutils.py", line 25, in <module>
    import netifaces
ImportError: No module named netifaces

PR #217, opened 2015-12-07
Fix missing dependency (netifaces).

mbookman:
Should fix: https://github.com/gc3-uzh-ch/elasticluster/issues/216


riccardomurri:
Thanks, merged!

PR #218, opened 2015-12-09
Update cluster.update() to try to set the node's preferred_ip

mbookman:
after calling node.update_ips()

I have occasionally had nodes in the saved nodes list which had no IP addresses saved.
Running "elasticluster setup" would call node.update_ips(), but then didn't try to "connect()" before handing the node to ansible (which then had no IP address to connect to).

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -769,6 +769,14 @@ class Cluster(Struct):
         for node in self.get_all_nodes():
             try:
                 node.update_ips()
+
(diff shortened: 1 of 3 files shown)


mbookman:
Added a somewhat related change. In this case, I found that some nodes had been deleted without Elasticluster being updated. Those nodes could then never be removed from the cluster as stop_instance would throw an exception.

riccardomurri:
Thanks, merged!

PR #243, opened 2016-04-19
Make ec2 access_key and secrets optional

jsharpe:
This allows for boto credentials to be passed via iam roles when running on ec2 or via a .aws/credentials file from the awscli tool rather than being explicitly placed in a configuration file.


riccardomurri:
Merged, thanks!

PR #260, opened 2016-06-01
Documentation: Use ec2_boto consistently to avoid confusion

pinguinkiste:
Both "boto" and "ec2_boto" are used inconsistently in the documentation and lead to some debugging because the wrong key was used (and accepted) in the config file. This patch makes the use consistent by changing all occurrences of "boto" to "ec2_boto".


riccardomurri:
Merged, thanks!

PR #261, opened 2016-06-01
Improve default config handling

pinguinkiste:
This change prevents a default config file from being created when config.d already exists, for example because of copying a different template.

The second commit fixes some PEP-8 violations and re-factors some code (high cyclomatic complexity, etc.) to make IntelliJ IDEA happy. ;) I'm not much of a Python person. The changes seem sensible (from a JVM-based/biased perspective), but please feel free to drop it if you disagree. (I only moved code, no functional changes.)

Best regards,

Manuel

### DIFF
--- a/elasticluster/main.py
+++ b/elasticluster/main.py
@@ -150,8 +150,8 @@ class ElastiCluster(cli.app.CommandLineApp):
                                  "%s\n" % (str(ex)))
                 sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


riccardomurri:
Can you please split the two commits into separate pull requests? They look pretty unrelated...

pinguinkiste:
Sure, will do.

PR #262, opened 2016-06-01
Do not create default config if config.d directory exists

pinguinkiste:


pinguinkiste:
I'll send the second pull request if this is merged, as there will be conflicts when I base both of them on master.

PR #263, opened 2016-06-02
Fix PEP-8 violations in elasticluster/main.py

pinguinkiste:
Second commit from yesterday. No code changes, just re-arrangements.

### DIFF
--- a/elasticluster/main.py
+++ b/elasticluster/main.py
@@ -15,7 +15,6 @@
 #   You should have received a copy of the GNU General Public License
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-__author__ = 'Nicolas Baer <nicolas.baer@uzh.ch>, Antonio Messina <antonio.s.messina@gmail.com>'
 
 # System imports
 import logging
@@ -32,12 +31,12 @@ try:
     from voluptuous import MultipleInvalid, Invalid
 except ImportError:
     # Voluptuous version <= 0.7.2
+    # noinspection PyUnresolvedReferences
     from voluptuous.voluptuous import MultipleInvalid, Invalid
 
 import coloredlogs
 
 from pkg_resources import resource_filename
-
 
 # Elasticluster imports
 from elasticluster import log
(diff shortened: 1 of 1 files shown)


elasticluster/main.py
```diff
@@ -15,7 +15,6 @@
 #   You should have received a copy of the GNU General Public License
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-__author__ = 'Nicolas Baer <nicolas.baer@uzh.ch>, Antonio Messina <antonio.s.messina@gmail.com>'
```
riccardomurri: Why moving this line down, after the imports?  As far as I can tell, it is customary in Python files to have such metadata at the beginning of the file. Does PEP-8 mandate a different convention?

elasticluster/main.py
```diff
@@ -32,13 +31,13 @@
     from voluptuous import MultipleInvalid, Invalid
 except ImportError:
     # Voluptuous version <= 0.7.2
+    # noinspection PyUnresolvedReferences
```
riccardomurri: I understand that these comments are inserted by some tool that you're using.  Is it possible to configure it to append a prefix/suffix indicating that they are not to be read by humans?  E.g.::

        # FooChecker: noinspection PyUnresolvedReferences

elasticluster/main.py
```diff
@@ -32,13 +31,13 @@
...
     from voluptuous.voluptuous import MultipleInvalid, Invalid
 
 import coloredlogs
 
 from pkg_resources import resource_filename
 
-
```
riccardomurri: I actually like to have 2 blank lines separating stdlib imports / 3rd party imports / local pkg imports.

(Yes, I know PEP-8 mandates otherwise but I find this convention more readable. I do not have strong opinion on it, though.)

riccardomurri:
LGTM, except for a couple minor points (see line notes)

elasticluster/main.py
pinguinkiste: Apparently, it seems that it mandates that all meta data should go after the imports, as this is where the file begins and people start reading. I've no hard feeling either way, as this is not my reasoning. I was quite surprised as well.

elasticluster/main.py
pinguinkiste: It's IntelliJ IDEA/PyCharm. It should be possible. I'll try and update the commit.

elasticluster/main.py
pinguinkiste: Me neither, I can revert that change. As said, not much of a Python person, just did what's supposed to be best practices. ;)

elasticluster/main.py
pinguinkiste: Does not work, it needs to be the line to be exactly like this. Both pre-pending and appending does not work. We can also drop it if you like.

riccardomurri:
Cherry-picked and adapted into commit 67c9518, now in master.

Thanks for contributing!

PR #264, opened 2016-06-02
.gitignore entries for IntelliJ IDEA/PyCharm

pinguinkiste:
Template from .ignore plug-in in IDEA with some modifications at the end (because I'm not sure if your OK with some IDEA-specific files in the repository).

Best regards,

Manuel

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,54 @@ setuptools-*.egg
 *.tar.bz2
 *.tar.gz
 *.zip
+
+### JetBrains template
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and Webstorm
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
+# User-specific stuff:
+.idea/workspace.xml
+.idea/tasks.xml
+.idea/dictionaries
+.idea/vcs.xml
+.idea/jsLibraryMappings.xml
+
+# Sensitive or high-churn files:
+.idea/dataSources.ids
+.idea/dataSources.xml
+.idea/dataSources.local.xml
+.idea/sqlDataSources.xml
+.idea/dynamic.xml
+.idea/uiDesigner.xml
+
+# Gradle:
+.idea/gradle.xml
+.idea/libraries
+
+# Mongo Explorer plugin:
+.idea/mongoSettings.xml
+
+## File-based project format:
+*.iws
+
+## Plugin-specific files:
+
+# IntelliJ
+/out/
+
+# mpeltonen/sbt-idea plugin
+.idea_modules/
+
+# JIRA plugin
+atlassian-ide-plugin.xml
+
+# Crashlytics plugin (for Android Studio and IntelliJ)
+com_crashlytics_export_strings.xml
+crashlytics.properties
+crashlytics-build.properties
+fabric.properties
+
+# Ignore everything in the .idea directory and the .iml for now. This can be removed later, since it will make sharing
+# of configuration easier if more developers use IDEA/PyCharm.
+.idea/*
+*.iml


.gitignore
```diff
@@ -44,3 +44,54 @@ setuptools-*.egg
...
+
+# Sensitive or high-churn files:
+.idea/dataSources.ids
+.idea/dataSources.xml
+.idea/dataSources.local.xml
+.idea/sqlDataSources.xml
+.idea/dynamic.xml
+.idea/uiDesigner.xml
+
+# Gradle:
```
riccardomurri: I'm pretty sure there's no role for Gradle in ElastiCluster :-)

.gitignore
```diff
@@ -44,3 +44,54 @@ setuptools-*.egg
...
+.idea/gradle.xml
+.idea/libraries
+
+# Mongo Explorer plugin:
```
riccardomurri: Likewise, why do we need to exclude Mongo stuff here?

.gitignore
```diff
@@ -44,3 +44,54 @@ setuptools-*.egg
...
+
+## Plugin-specific files:
+
+# IntelliJ
+/out/
+
+# mpeltonen/sbt-idea plugin
+.idea_modules/
+
+# JIRA plugin
```
riccardomurri: Same as above: not used in ElastiCluster -- why bother?

.gitignore
```diff
@@ -44,3 +44,54 @@ setuptools-*.egg
...
+atlassian-ide-plugin.xml
+
+# Crashlytics plugin (for Android Studio and IntelliJ)
```
riccardomurri: Same as above: not used in ElastiCluster -- why bother?

.gitignore
```diff
@@ -44,3 +44,54 @@ setuptools-*.egg
...
+com_crashlytics_export_strings.xml
+crashlytics.properties
+crashlytics-build.properties
+fabric.properties
+
+# Ignore everything in the .idea directory and the .iml for now. This can be removed later, since it will make sharing
+# of configuration easier if more developers use IDEA/PyCharm.
+.idea/*
```
riccardomurri: Why go exclude IDEA files one by one, if you're going to exclude the whole directory in the end? Just keep this line, and forget the rest...

riccardomurri:
I have no objections including exclusion patterns for IDEA-generated files, but the list looks to me unnecessarily redundant -- I would just keep the exclusion of the `.idea/` and `.idea_modules/` directory, and perhaps `*.iml`/`*.iws` files (if they are generated outside those two dirs).  

Feel free to correct me, though - I do not use PyCharm so I'm mostly guessing here.

.gitignore
```diff
@@ -44,3 +44,54 @@ setuptools-*.egg
...
+# Mongo Explorer plugin:
+.idea/mongoSettings.xml
+
+## File-based project format:
+*.iws
+
+## Plugin-specific files:
+
+# IntelliJ
+/out/
```
riccardomurri: Exclude patterns should be relative to the Git repo root

.gitignore
pinguinkiste: I excluded the directory in the end because I thought you might not be comfortable with IDEA-specific files in the directory. It can make sense to share some of them, so people share some setting, e.g. formatter settings or such things. This is why only some files are excluded in the default pattern. I added the extra exclude because of that. Alternatively, I could add those files (defaults) to the repository.

.gitignore
pinguinkiste: Agreed. It's from the template. My take on this is usually that these do not hurt and make update to the template easier to merge. It's obviously unnecessary. I can strip it down if you like your .gitignore short. No hard feelings either way.

.gitignore
pinguinkiste: I just tried it, it seems like it is sufficient to have just ".idea/" in the file to cover everything if a recent version of IDEA is used.

riccardomurri:
I'm fine merging -- would like to keep the `.gitignore` file short and tidy but no strong opinion: it's basically a write-only file.

I'll trust your judgement re: what exclusion patterns to use for IDEA/PyCharm, I'm not using either one, as said.

Once you update and give me a go, I'll merge.

.gitignore
riccardomurri: So should I just add that line to `.gitignore` and discard the rest of this PR?

riccardomurri:
Cherry-picked and merged in commit 2a65004

Thanks for contributing!

PR #265, opened 2016-06-02
Documentation: Explain that different flavors can be used on compute …

### LINKED ISSUE #245: Different image types for frontend and compute nodes
I should be able to specify different node types for frontend and compute nodes as I may want a small e.g. t2.micro frontend and large compute nodes.

pinguinkiste:
…and frontend nodes

Fixes #245.

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -583,7 +583,9 @@ Mandatory configuration keys
 
     the image type to use. Different cloud providers call it
     differently, could be `instance type`, `instance size` or
-    `flavor`.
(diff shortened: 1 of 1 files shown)


riccardomurri:
Merged, thank you very much for all these improvements!

pinguinkiste:
You're welcome! There may be more to come... ;) We're evaluating ElastiCluster for use at our institution, so I just took some time to start to work with the code while reading it or testing ElastiCluster.

Best,

Manuel

PR #269, opened 2016-06-08
Fetch hostkeys securely from launched instances

manuelprinz:
Closes #144.

I took the liberty to update the code from @jwm to work on `master`, although there are some slight changes. I tried to take your feedback into account. I also implemented the option for OpenStack. I have no way to test EC2 at the moment, so I just included the changes to `ec2_boto` as they were in @jwm 's patch.

It works on our OpenStack instance, further testing and feedback welcome, as I'm still not all to deep into the code.

Best,

Manuel

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -670,6 +670,11 @@ Optional configuration keys
     Only supported when the cloud provider is `google`.
     Only supported value is `preemptible`.
 
+``ssh_hostkeys_from_console_output``
+
+    whether to retrieve SSH host keys from instances´ console output,
+    instead of accepting the host keys without verification when connecting
+    via SSH for the first time.
 
 Examples
 --------
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -45,12 +45,6 @@ from elasticluster.exceptions import TimeoutError, NodeNotFound, \
 from elasticluster.repository import MemRepository
 
 
-class IgnorePolicy(paramiko.MissingHostKeyPolicy):
-    def missing_host_key(self, client, hostname, key):
-        log.info('Ignoring unknown %s host key for %s: %s' %
-                 (key.get_name(), hostname, hexlify(key.get_fingerprint())))
-
-
 class Struct(object, UserDict.DictMixin):
     """
     This class is a clone of gc3libs.utils.Struct class from GC3Pie project: https://code.google.com/p/gc3pie/
@@ -151,7 +145,7 @@ class Cluster(Struct):
     :ivar nodes: dict [node_type] = [:py:class:`Node`] that represents all
                  nodes in this cluster
     """
-    startup_timeout = 60 * 10  #: timeout in seconds to start all nodes
+    startup_timeout = 60 * 15  #: timeout in seconds to start all nodes
 
 
     def __init__(self, name, user_key_name='elasticluster-key',
@@ -175,6 +169,8 @@ class Cluster(Struct):
 
         self.user_key_public = os.path.expanduser(user_key_public)
         self.user_key_public = os.path.expandvars(user_key_public)
+
+        self.ssh_hostkeys_from_console_output = extra.get('ssh_hostkeys_from_console_output')
 
         # this needs to exist before `add_node()` is called
         self._naming_policy = NodeNamingPolicy()
@@ -573,6 +569,10 @@ class Cluster(Struct):
             while pending_nodes:
                 for node in pending_nodes[:]:
                     ssh = node.connect(keyfile=self.known_hosts_file)
+                    if self.ssh_hostkeys_from_console_output:
+                        self.get_ssh_key_from_console_output(keys, node.instance_id, node.ips)
+
+                    ssh = node.connect(keyfile=self.known_hosts_file)
                     if ssh:
                         log.info("Connection to node %s (%s) successful.",
                                  node.name, node.connection_ip())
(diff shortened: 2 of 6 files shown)


elasticluster/cluster.py
```diff
@@ -151,7 +145,7 @@ class Cluster(Struct):
     :ivar nodes: dict [node_type] = [:py:class:`Node`] that represents all
                  nodes in this cluster
     """
-    startup_timeout = 60 * 10  #: timeout in seconds to start all nodes
+    startup_timeout = 60 * 15  #: timeout in seconds to start all nodes
```
riccardomurri: Are you increasing the timeout because the machine needs to have finished booting before we can process the host keys?

elasticluster/cluster.py
```diff
@@ -829,6 +829,20 @@ def update(self):
                           node, str(ex))
         self.repository.save_or_update(self)
 
+    def get_ssh_key_from_console_output(self, known_hosts, instance_id, names):
+        console_output = self._cloud_provider.get_console_output(instance_id)
+        hostkeys = re.sub(
+            r'^.*-+BEGIN SSH HOST KEY KEYS-+\s*(.*)[\r\n]-+END SSH HOST KEY KEYS-+.*',
```
riccardomurri: Unless there is evidence that the number of hyphens is varying, I think it would be safer to hard-code it (i.e., `---BEGIN SSH ...` instead of `-+ BEGIN SSH ...`)

elasticluster/cluster.py
```diff
@@ -829,6 +829,20 @@ def update(self):
                           node, str(ex))
         self.repository.save_or_update(self)
 
+    def get_ssh_key_from_console_output(self, known_hosts, instance_id, names):
+        console_output = self._cloud_provider.get_console_output(instance_id)
+        hostkeys = re.sub(
```
riccardomurri: I would rather do:

    # outside method, at class level:
    HOSTKEYS_RE = re.compile('---BEGIN ...', re.DOTALL)

    # line XX and following:
    match = self.HOSTKEYS_RE.search(console_output)
    if match:
      hostkeys = match.group(1)
      # ... rest of hostkeys code here

The reasons are:

- it's slightly faster (no string substitution -- just extract the substring you need, plus the regex is compiled)
- one can handle the case when the keys are not found in the console output

elasticluster/conf.py
```diff
@@ -451,6 +451,7 @@ def validate(self):
         schema = {"cluster": {"cloud": All(str, Length(min=1)),
                               "setup_provider": All(str, Length(min=1)),
                               "login": All(str, Length(min=1)),
+                              Optional("ssh_hostkeys_from_console_output"): Boolean(str),
```
riccardomurri: Why is an explicit option needed? Can't you just do it and if it fails, fall back to the current behavior?

docs/configure.rst
```diff
@@ -670,6 +670,11 @@ Optional configuration keys
     Only supported when the cloud provider is `google`.
     Only supported value is `preemptible`.
 
+``ssh_hostkeys_from_console_output``
```
riccardomurri: Why an explicit option? Can't you just try and fall back on the current behavior on failure?

elasticluster/cluster.py
```diff
@@ -45,12 +45,6 @@
 from elasticluster.repository import MemRepository
 
 
-class IgnorePolicy(paramiko.MissingHostKeyPolicy):
```
riccardomurri: This is still needed for a fall-back mechanism.

(Which we need to keep until it's sure that also GCE and Azure also print the SSH keys to the console.)

elasticluster/cluster.py
```diff
@@ -176,6 +170,8 @@ def __init__(self, name, user_key_name='elasticluster-key',
         self.user_key_public = os.path.expanduser(user_key_public)
         self.user_key_public = os.path.expandvars(user_key_public)
 
+        self.ssh_hostkeys_from_console_output = extra.get('ssh_hostkeys_from_console_output')
```
riccardomurri: See above re: explicit option.

riccardomurri:
Thanks a lot for contributing!

The code looks generally good and I'm willing to merge but there are a couple of points that need to be addressed before:

1. The current behavior has to stay as a fall-back mechanism: I'm not sure if the console trick works on GCE and I don't want to break it.

2. If we have a fall-back, then there's little sense in having an explitcit option to enable parsing the console output: just try it, and fall-back to the insecure behavior on failure.

I'll be on vacation in the next two weeks, so I will not merge until I come back (it's a risky part of the code ;-)).

elasticluster/cluster.py
pinguinkiste: It was changed in the previous PR and not commented by anyone, so I kept it. It does not seem to make a difference at all, so I'll just revert it.

elasticluster/cluster.py
pinguinkiste: Agreed, also taken from previous PR.

elasticluster/conf.py
pinguinkiste: Taken from original PR. I'm happy to change it in the way you think is most reasonable.

elasticluster/cluster.py
pinguinkiste: Taken from previous PR. This was shown to be unused by my IDE and I did not find any instances of it in the code, so assumed it to be safe to keep this change. Where/When is it used? What did I overlook?

pinguinkiste:
No worries! I'll change the code as you propose. I just took what was available to get rid of the crude hack we use locally at the moment. ;-) This will also give me some time to test it properly.

Enjoy your holiday!

Best,

Manuel

PR #270, opened 2016-06-08
Wait until OpenStack instance provisioning is finished

pinguinkiste:
A small change (and tests) that make ElastiCluster wait until OpenStack instances have finished setting up. It assumes that `cloud-init` is used which should be the case on all OpenStack instances. That said, I'm not sure how much we should rely on it. (But everyone I asked said it is a reasonable assumption to expect `cloud-init` to be present.)

### DIFF
--- a/elasticluster/providers/openstack.py
+++ b/elasticluster/providers/openstack.py
@@ -19,6 +19,7 @@
 # You should have received a copy of the GNU General Public License along
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+import re
 
 __docformat__ = 'reStructuredText'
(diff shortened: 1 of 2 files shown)


riccardomurri:
LGTM, but why? there's quite some code that deals with waiting for an instance to be able to respond to SSH connections later on... It is provider-independent hence makes no assumptions like the use of `cloud-init`.

pinguinkiste:
We had some issues with that. ElastiCluster tried to continue before the host was properly set up and failed. This fixed this kind of "race condition" for us.

However, I kind of developed this and the previous PR together and split them later. Maybe there was some interdependent bug. I'll do some more test without this change and see what happens. If this works out alright, I'll withdraw this one.

pinguinkiste:
After some testing it seem that this is not needed. Closing. Sorry for the noise!

PR #271, opened 2016-06-09
Save SSH host keys of contacted hosts

pinguinkiste:
Fixes issue #144.

While updating the (other) PR for #144 I noticed that the code for fetching the keys from the console output did not get called. I tried to dig into it and it seems that the keys where simply "outdated". I made this minimal patch that fixes the problems for me. I'll leave the other PR open until it is decided if this is sufficient.

Best regards,

Manuel

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -581,6 +581,7 @@ class Cluster(Struct):
                             for ktype, keydata in key.items():
                                 keys.add(host, ktype, keydata)
                         pending_nodes.remove(node)
+                    self._save_keys_to_known_hosts_file(keys)
                 if pending_nodes:
                     time.sleep(5)
 
@@ -602,10 +603,7 @@ class Cluster(Struct):
         self.repository.save_or_update(self)
 
         # Save host keys
-        try:
-            keys.save(self.known_hosts_file)
-        except IOError:
-            log.warning("Ignoring error while saving known_hosts file %s" % self.known_hosts_file)
+        self._save_keys_to_known_hosts_file(keys)
 
         # A lot of things could go wrong when starting the cluster. To
         # ensure a stable cluster fitting the needs of the user in terms of
@@ -622,6 +620,12 @@ class Cluster(Struct):
                     min_nodes[group] = len(nodes)
 
         self._check_cluster_size(min_nodes)
+
+    def _save_keys_to_known_hosts_file(self, keys):
+        try:
+            keys.save(self.known_hosts_file)
+        except IOError:
+            log.warning("Ignoring error while saving known_hosts file %s" % self.known_hosts_file)
 
     def _check_cluster_size(self, min_nodes):
         """Checks the size of the cluster to fit the needs of the user. It


elasticluster/cluster.py
```diff
@@ -572,6 +572,7 @@ def timeout_handler(signum, frame):
         try:
             while pending_nodes:
                 for node in pending_nodes[:]:
+                    self._save_keys_to_known_hosts_file(keys)
```
riccardomurri: Shouldn't this be done *after* the node has been contacted?  I.e., between lines 584 and 585

Otherwise the host keys for the last node in the list will not be saved...

riccardomurri:
LGTM, with a couple of questions:

1. Shouldn't the save in the "pending nodes" loop happen later? See line comment

2. How does this fix #144?  My understanding is that PR #269 fixes #144, while *this* PR is pretty independent of it...

elasticluster/cluster.py
pinguinkiste: Yes, you're right. Didn't happen when testing it but moving down is safer.

pinguinkiste:
On 2.: The current situation is that on current `master` provisioning with OpenStack does not work. It's trapped in an endless `Invalid host key: host X (IP); check keyfile:` cycle until time-out, although everything is fine. This PR fixes this problem. PR #269 does not seem to fix the problem without saving the keys as well. If I save them, the code extracting them never gets called. So the minimum change to make ElastiCluster exit the unless loop is this one.

So I'd not say it's "independent", rather "not the solution we're looking for", maybe... Depends on if the keys should be fetched from the console (as a feature) or just need to be fetched somehow in order to continue. This only fixes the bug, it does not implement the feature.

I'll look at that more deeply, as I am confused. (It's been a long and hard week.)

pinguinkiste:
I changed the PR you proposed above and tested it. I also changed the title and removed the reference to #144 so it will not be closed if you decide to merge it.

riccardomurri:
> The current situation is that on current master provisioning with OpenStack does not work.

Well, we provision with OpenStack all the time (it's the only in-house infrastructure) so I'm pretty sure it *does* work :-)

I would be curious about what's different in your OpenStack w.r.t. ours that makes the current code not work...

riccardomurri:
Anyway: looks good so merging.  Thank you for contributing!

pinguinkiste:
I didn't mean to imply that it doesn't work anywhere and did not wan to offend you. I am sorry!

The only thing I can say about our installation is that it is Juno. For any other data you might need I'd be happy to ask the people who maintain it. ;-)

riccardomurri:
No offence taken! :-) I am sorry if I sounded offended or harsh!

riccardomurri:
Thinking about it, I don't think OpenStack can play any role in this -- if you are able to contact the SSH server, then the VM is up and the security groups allow tcp port 22.

So I'd look for differences in the client-side SSH config.  Do you have any options enabled that mandate strict host key checking or that a host be known in advance?

PR #283, opened 2016-06-21
Install MAUI scheduler from RPM

pinguinkiste:
MAUI is not available in any of the installed repositories anymore, so it's fetched
as an RPM directly. The package provides it as "maui.d" so services calls are updated
accordingly.

### DIFF
--- a/elasticluster/share/playbooks/roles/pbs+maui/tasks/maui.yml
+++ b/elasticluster/share/playbooks/roles/pbs+maui/tasks/maui.yml
@@ -1,9 +1,7 @@
 ---
-- name: Install MAUI package
+- name: Install MAUI RPM from a remote repo
   when: is_centos
-  action: yum pkg={{item}} state=latest
-  with_items:
-    - maui
+  yum: name=http://ftp.cs.stanford.edu/pub/rpms/centos/6.2/x86_64/maui-3.3.1-x86_64-fpmbuild.rpm state=present
 
 - name: Ensure /var/spool/maui/log directory exists
   action: file path=/var/spool/maui/log state=directory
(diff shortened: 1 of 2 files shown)


elasticluster/share/playbooks/roles/pbs+maui/tasks/maui.yml
```diff
@@ -1,9 +1,7 @@
 ---
-- name: Install MAUI package
+- name: Install MAUI RPM from a remote repo
   when: is_centos
-  action: yum pkg={{item}} state=latest
-  with_items:
-    - maui
+  yum: name=http://ftp.cs.stanford.edu/pub/rpms/centos/6.2/x86_64/maui-3.3.1-x86_64-fpmbuild.rpm state=present
```
riccardomurri: Would it make sens to substitute `/{{ansible_distribution_major_version}}/` for `/6.2/`?

That would make the package URL work on different/updated RHEL/CentOS versions, at least in theory...

riccardomurri:
LGTM.

One more small request: can you also please edit the docs that currently state that you have to provide you own RPMs?

As always, thanks for contributing!

chrifey:
One could also compile maui, but look like at least a valid login on adaptiveomputing.com is needed, so no mode.

pinguinkiste:
Indeed. Riccardo, how should we go about that? My proposal would be to leave the change as is but add a comment that it also works fine for CentOS 7. We have no influence on the upstream repository, so I see very little chance that they will fix it. I can ask, though.

riccardomurri:
Merged as ffd39b8 -- thanks for contributing!

PR #284, opened 2016-06-21
Add NUMA workaround for PBS

pinguinkiste:
PBS does not work correctly on a single compute node with NUMA support. More
information can be found at [https://bugzilla.redhat.com/show_bug.cgi?id=1321154](https://bugzilla.redhat.com/show_bug.cgi?id=1321154).

### DIFF
--- a/elasticluster/share/playbooks/roles/pbs+maui/templates/var/spool/torque/server_priv/nodes.j2
+++ b/elasticluster/share/playbooks/roles/pbs+maui/templates/var/spool/torque/server_priv/nodes.j2
@@ -1,3 +1,3 @@
 {% for host in groups['pbs_clients'] %}
-{{ host }} np={{ hostvars[host].ansible_processor_count }}
+{{ host }} np={{ hostvars[host].ansible_processor_count }} num_node_boards=1
 {% endfor %}


riccardomurri:
Hmmm followup comments to the link you posted state that the workaround has it flaws too: https://bugzilla.redhat.com/show_bug.cgi?id=1321154#c37

pinguinkiste:
Yes, it is not perfect. However, without this workaround a cluster does not work at all, so there is no point in having one if you cannot submit jobs to it. We do not experience the other problems described in the bug, but they seem to be there.

I have no idea how many people use ElastiCluster with PBS, so I cannot say how this affects others. I do not mind maintaining this change locally if you think it should not be included. Your call. ;-)

riccardomurri:
Merged, thanks!

riccardomurri:
I have made a little edit in ad455b0 to apply this workaround only when a node has more than 1 processor.  Can you please check that it (still) works?

pinguinkiste:
Hi Riccardo,

sorry for the late reply, first came the holiday (although very short) and then the resulting backlog... ;-)

I'll try out your changes and report back. If I do not reopen, consider it working fine.

Thanks for merging all the other patches!

PR #285, opened 2016-06-21
Change paths of Torque from /var/spool to /var/lib

manuelprinz:
It seems there was an update in the latest Torque packages that make the
Ansible playbooks fail as things are expected in the wrong places.


riccardomurri:
LGTM, but can you please include a comment or note as to what package version/release this applies to?  Otherwise playbooks will fail the other way round if somebody tries to install an older version of the SW (e.g., because they have an older version of the base OS)

pinguinkiste:
Sure. What do you think the best place would be to do that? In the documentation or in all files affected by this change?

riccardomurri:
Well, at least in `docs/playbooks.rst`, in the PBS/TORQUE section.

Depending on applicability of tha patch (TORQUE version, OS distro and version), we might need to add a conditional in the Ansible playbooks to cater for both locations.

(I can do that myself, but I need to know what exact version of TORQUE and OS this applies to... Debian and Ubuntu, which I follow most closely, still ship with TORQUE 2.4.16)

riccardomurri:
Eventually managed as commit 4bdce67 -- needed a rebase to apply cleanly to current ElastiCluster sources.

Many many thanks for contributing to ElastiCluster!

PR #286, opened 2016-06-21
Fix YAML syntax errors in playbooks for PBS

pinguinkiste:


elasticluster/share/playbooks/roles/nfs-server/defaults/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 
 NFS_EXPORTS:
-  - path: '/home'
+    path: '/home'
```
riccardomurri: Well, no: `NFS_EXPORTS` should indeed be a *list* (each item of which is a dictionary with keys `path`, `clients`, `options`)

riccardomurri:
The PBS/MAUI playbooks edits look good (thx for fixing!), but the edit on `roles/nfs-server` does not. (Unless I am missing something?)

chrifey:
Hi, from our experience the playbook has to look like the following (example shows, that there may be multiple export paths, and therefore the "-path" is a list:

```yml
 - role: nfs-server
      NFS_EXPORTS:
      - path: '/home'
        clients: '{{groups.pbs_clients}}'
        options: 'rw,no_root_squash,async'
      - path: '/tmp'
        clients: '{{groups.pbs_clients}}'
        options: 'rw,no_root_squash,async'
```

That works and generated the following exports:

```bash
[root@frontend001 ~]# cat /etc/exports
/home compute001(rw,no_root_squash,async)
/tmp compute001(rw,no_root_squash,async)
```

If we do not insert the dash the task fails with the following error:

```bash
TASK [nfs-server : Export directories] *****************************************
task path: /root/.python-eggs/elasticluster-1.3.dev0-py2.7.egg-tmp/elasticluster/share/playbooks/roles/nfs-server/tasks/main.yml:43
fatal: [frontend001]: FAILED! => {"failed": true, "msg": "'unicode object' has no attribute 'path'"}```

riccardomurri:
Merged into "master" as e987dab -- thanks a lot for contributing!

PR #287, opened 2016-06-21
Apply dynamic include workaround for Ansible playbooks

manuelprinz:
There seems to be regression in Ansible that breaks dynamic includes with
every few minor releases. This workaround turns the includes into static
includes which work across all versions. More information can be found in
this Google Groups discussion:

 https://groups.google.com/forum/#!msg/ansible-devel/9aJaoVeRdOg/B4TvRTLgCAAJ

### DIFF
--- a/elasticluster/share/playbooks/roles/pbs+maui.yml
+++ b/elasticluster/share/playbooks/roles/pbs+maui.yml
@@ -14,18 +14,24 @@
         options: 'rw,no_root_squash,async'
   tasks: 
     - include: pbs+maui/tasks/master.yml
+      static: yes
     - include: cluster/tasks/packages.yml
+      static: yes
   handlers:
     - include: slurm/handlers/main.yml
     - include: common/handlers/main.yml
+      static: yes
     - include: pbs+maui/handlers/main.yml
+      static: yes
 
 - name: MAUI masternode Playbook
   hosts: maui_master
   tasks: 
     - include: pbs+maui/tasks/maui.yml
+      static: yes
   handlers:
     - include: pbs+maui/handlers/main.yml
+      static: yes
 
 - name: PBS worker nodes Playbook
   hosts: pbs_clients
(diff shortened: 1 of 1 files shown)


riccardomurri:
This is scary -- if Ansible changed the `include:` behavior in a way that breaks ElastiCluster's use of it, there's a bootload of occurrences that need to be changed ...

riccardomurri:
The ["Dynamic vs Static includes"][1] section of the Ansible docs states that, starting Ansible 2.1, the following policy applies:

> By default in Ansible 2.1 and higher, includes are automatically treated as static rather than dynamic when the include meets the following conditions:
>
> * The include does not use any loops
> * The included file name does not use any variables
> * The static option is not explicitly disabled (ie. static: no)
> * The ansible.cfg options to force static includes (see below) are disabled

According to this policy, there should be no need to patch the playbooks as
`static: yes` would be the default behavior in all cases covered by this patch.
So maybe just requiring Ansible >= 2.1 would solve the issue?

Were you running Ansible 2.0 when you developed this PR?

There are, however, compatibility options that break the way `include:` is used
in ElastiCluster:


> Two options are available in the ansible.cfg configuration for static includes:
>
> * `task_includes_static` - forces all includes in tasks sections to be static.
> * `handler_includes_static` - forces all includes in handlers sections to be static.
>
> These options allow users to force playbooks to behave exactly as they did in 1.9.x and before.

If people normally use these, then I'm afraid I'll have to tag each `include:`
statemement with an accompanying `static: yes|no` line. (Also assuming the two
configuration options below do not override what's in the playbook.)

[1]: http://docs.ansible.com/ansible/playbooks_roles.html#dynamic-versus-static-includes

riccardomurri:
I think this is no longer relevant in the current version of Ansible, which now uses different "import" and "include" keywords for the static and dynamic behavior.

At any rate, I think that ElastiCluster's current "master" branch does not suffer from static/dynamic import issues any more.

Many thanks for taking the time to dig into the issue and contribute!

Issue #289, opened 2016-06-29
Slurm + GlusterFS


GlastonburyC:
I would like a cluster using slurm with Gluster.

This is because I'm coming into a read/write bottleneck with a simple NFS mounted volume. I have included an excerpt from my config file below to see if it's correct?

```
[cluster/slurm+gluster]
cloud=openstack
network_ids=7d326b6a-fae5-4c4f-87e4-e7b5a3628fc1
login=centos
setup_provider=ansible-slurm,gluster
security_group=icmp-ssh-rna-seq-map-ref
image_id=1b887b94-bbcc-4736-a92c-e54c7c56310f
flavor=CraigG
frontend_nodes=1
compute_nodes=10
ssh_to=frontend
```
```
[setup/ansible-slurm,gluster]
provider=ansible
frontend_groups=slurm_master,glusterfs_client,glusterfs_server
compute_groups=slurm_workers,glusterfs_client
ansible_ssh_pipelining=no
```

I get the error:

```
TASK [slurm-master : Create cluster in accounting database] ********************
[...]

PR #290, opened 2016-06-30
Variable was not passed into include (was working prior to version 2.…

ewiger:
…0 of ansible). Fixed by adding support for loop control (Also see http://docs.ansible.com/ansible/playbooks_loops.html#loops-and-includes)

### DIFF
--- a/elasticluster/share/playbooks/roles/glusterfs-server/tasks/main.yml
+++ b/elasticluster/share/playbooks/roles/glusterfs-server/tasks/main.yml
@@ -22,7 +22,7 @@
     - gluster
     - glusterfs-server
 
-    
+
 - name: Ensure the GlusterFS service is running (RHEL-compatible)
   service:
     name='{{item}}'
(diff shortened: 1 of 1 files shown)


riccardomurri:
Is this related to the infinite loop/expansion of `item` reported in #289 ?

However the docs state that:

- `loop_control` is only available in Ansible 2.1+
- `item` is the default name for the loop variable in Ansible 2.x

So it would be sufficient to just replace `include: export.yml item='{{item}}'` with `include: export.yml` ?  Otherwise we would also have to bump the dependency on Ansible to version 2.1 ...

GlastonburyC:
Thanks for looking at this. if you could tell me when i can pull the latest version?

riccardomurri:
I think the infinite loop is fixed in commit e4697b3 -- @ewiger can you please confirm?

riccardomurri:
I cannot reproduce this with the latest "master" code; assuming it's fixed.

Issue #267, opened 2016-06-03
Broken on CentOS 7 due to requiretty in `/etc/sudoers`


pinguinkiste:
I tried to get ElastiCluster run with CentOS 7 images. It fails in the setup task because CentOS has `Defaults requiretty` in `/etc/sudoers` but ansible does not run with a TTY (obviously). This is with `image_sudo=True` in the config. If I change it to `image_sudo=False` it runs the first task but fails when adding the EPEL repository with: `fatal: [frontend001]: FAILED! => {"changed": true, "failed": true, "msg": "You need to be root to perform this command.\n", "rc": 1, "results": ["Loaded plugins: fastestmirror\n"]}`

All of my attempts to work around that (by removing the line from `/etc/sudoers` have failed so far. Any pointers on how to fix this would be great.

Best,

Manuel

verdurin:
@pinguinkiste - see #183 - there is a workaround

pinguinkiste:
Thanks @verdurin! I wonder why the issue did not turn up in my search!? Anyway, setting the environment variable works. Closing this issue.

Issue #183, opened 2015-09-24
Sudo reconfiguration needed for CentOS nodes


verdurin:
Ansible playbooks all fail with CentOS nodes (at least when trying to build a SLURM cluster) because the `requiretty` setting in sudo isn't disabled.

See

https://bugzilla.redhat.com/show_bug.cgi?id=1020147

riccardomurri:
This should now work with code from the branch `feature/forward-compatibility-with-ansible` (soon to be merged into `develop`), provided you turn "SSH pipelining" off:

        env ANSIBLE_SSH_PIPELINING=no elasticluster setup my-cluster # ...

verdurin:
Was this ever merged? I can't see the branch, nor any indication of this fix in the `develop` log.

riccardomurri:
As far as I can recall, the problem just is no longere there if you use Ansible >= 2.0 (which in turn requires ElastiCluster "develop")

verdurin:
Just tried with the `develop` branch, and that still has the check for Ansible < 2.0...

ChrisHad:
I've been using the develop branch and I still get the sudo issue even with Ansible >= 2.0.
I tried CentOS 7 and 6. 
Is there a way to enable more verbose output from ansible/elasticluster? I tried adding -vvv and creating an alias to 'ansible-playbooks -v' but it doesn't seem to do anything.

riccardomurri:
In the "develop" branch, running `elasticluster -vvv` does actually run `ansible-playbook -vvv` -- what branch are you running elasticluster from?

riccardomurri:
@verdurin The "develop" branch now requires Ansible >= 2.0 -- sorry I thought I had fixed this long ago so it went out of my radar ...

ChrisHad:
I'm running the develop branch:
```
# On branch develop
# Untracked files:
#   (use "git add <file>..." to include in what will be committed)
#
#       setuptools-8.0-py2.7.egg
#       setuptools-8.0.zip
nothing added to commit but untracked files present (use "git add" to track)

```

I compared the output with and without -vvv and it doesn't seem to make any difference.
But I'm only concerned about the verbose as I was trying to workaround the sudo issue.

riccardomurri:
Did you re-run `pip install -e .` after switching branches?

ChrisHad:
I installed straight from develop branch and with "_python setup.py install_". 
I also tried " _pip install -e ._" but no difference. 
[...]

PR #292, opened 2016-07-01
[WIP] Mask munge service from starting on install to fix permissions error

corburn:
I would appreciate any feedback regarding Ansible best practices or suggestions for a better approach to resolve these issues.

Issue 1a: Munge Daemon Logfile Security Check
-----------------------------------------------------------------

Ubuntu 15.04 switched to systemd and granted group write permission to `/var/log` for the group `syslog`.

The SLURM Munge service expects `/var/log/` to be writeable exclusively by either `root` or `munge`. As a result the munge service will raise the following error:

>   munged: Error: Logfile is insecure: group-writable permissions set on "/var/log"
     dun/munge/issues/31

Munge has fixed the permissions issue with commit https://github.com/dun/munge/commit/ec4a2c679d43db5e194434bdc7489e1a92599aa5, but previous versions must start the daemon with the `--syslog` flag to disable the error.

Issue 1b: Ubuntu/Debian Starts Daemons on Install
--------------------------------------------------------------------

Because Ubuntu starts the munge daemon on install, the task will fail before another task can append the `--syslog` flag munge.service.

Issue 2: Install dependencies
---------------------------------------

`openmpi-checkpoint` package was not found in the Ubuntu/Xenial repository

A sudo security error was triggered by the [requiretty](https://bugzilla.redhat.com/show_bug.cgi?id=1020147) option in the sudoers file while installing aptitude.

> sorry, you must have a tty to run sudo

The ansible apt module will use `aptitude` to install upgrades with the `upgrade=safe` flag, but `apt-get` with `upgrade=dist`.

Solution 1:
--------------

The pull request is [masking](https://fedoramagazine.org/systemd-masking-units/) the munge.service to prevent Ubuntu from starting the munge daemon on install followed by appending `--syslog` to the munge.service to resolve the permissions error.

"Solution 2":
----------------

Avoid installing the problematic dependencies because I wanted it to get it working and am not yet aware of a better solution. Removing the `requiretty` option from the sudoers file might resolve the sudo error, but given it is deliberately turned on by the distribution maintainers I am less inclined to disable it without understanding the implications.

### DIFF
--- a/elasticluster/share/playbooks/roles/common/tasks/deb.yml
+++ b/elasticluster/share/playbooks/roles/common/tasks/deb.yml
@@ -14,7 +14,7 @@
 
 - name: Upgrade all installed packages to latest version
   apt:
-    upgrade=safe
+    upgrade=dist
   when: is_debian_or_ubuntu
   
 - name: Ensure additional packages are installed
@@ -22,7 +22,7 @@
     name={{item}}
     state=present
   with_items:
-    - aptitude
+#    - aptitude
     - python-apt
     - sysvinit-utils
     - software-properties-common
--- a/elasticluster/share/playbooks/roles/slurm-common/tasks/main.yml
+++ b/elasticluster/share/playbooks/roles/slurm-common/tasks/main.yml
@@ -90,7 +90,7 @@
     - dmtcp
     - libopenmpi-dev
     - openmpi-bin
-    - openmpi-checkpoint
+    #- openmpi-checkpoint
   when: '{{is_debian_or_ubuntu}}'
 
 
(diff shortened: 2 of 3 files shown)


riccardomurri:
I think the MUNGE issue is fixed in commit 989841f (which basically cherry-picks your proposed edits on file `munge.yml`)

Likewise, the issue with `openmpi-checkpoint` is solved in the current "master"; C/R support was removed in Debian and Ubuntu packages citing lack of upstream support, so it doesn't make sense to install it anyway.

I cannot reproduce the issue with `requiretty` on Ubuntu; a similar issue exists on CentOS (see #267 and #183), and it is solved by turning off Ansible SSH pipelining (set the enviroment variable `ANSIBLE_SSH_PIPELINING=no`, or the equivalent setting in the ElastiCluster config file).

riccardomurri:
Do you think the solution outlined in the previous comment is satisfactory?  Can I close this PR?

PR #296, opened 2016-07-23
added support for ec2 spot instances

pvrancx:
Added basic support for ec2 spot instances.
Provide a price argument to the BotoProvider start_instance method to request a spot instance.
You can also provide an optional timeout (s) to determine how long the provider will wait for the request to be fulfilled.

I also added an example to the cluster config to show how a cluster can be launched with spot compute nodes (while using non-spot for the frontend )

### DIFF
--- a/elasticluster/providers/ec2_boto.py
+++ b/elasticluster/providers/ec2_boto.py
@@ -20,6 +20,7 @@ __author__ = 'Nicolas Baer <nicolas.baer@uzh.ch>, Antonio Messina <antonio.s.mes
 import os
 import urllib
 import threading
+import time
 from warnings import warn
 
 # External modules
@@ -153,6 +154,7 @@ class BotoCloudProvider(AbstractCloudProvider):
     def start_instance(self, key_name, public_key_path, private_key_path,
                        security_group, flavor, image_id, image_userdata,
                        username=None, node_name=None, network_ids=None,
+                       price=None,timeout=None,
                        **kwargs):
         """Starts a new instance on the cloud using the given properties.
         The following tasks are done to start an instance:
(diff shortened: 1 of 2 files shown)


elasticluster/providers/ec2_boto.py
```diff
@@ -206,18 +208,40 @@ def start_instance(self, key_name, public_key_path, private_key_path,
...
+            if price:
+                log.info("Requesting spot instance with price `%s`.",price)
+                request = connection.request_spot_instances(
+                                price,image_id, key_name=key_name, security_groups=security_groups,
+                                instance_type=flavor, user_data=image_userdata,
+                                network_interfaces=interfaces)[-1]
+                                
+                #wait until spot request is fullfilled (will wait forever if no timeout is given)
+                start_time = time.time()
+                log.info("Waiting for spot instance - timeout: `%s`.",str(timeout))
```
riccardomurri: It's not necessary to convert `timeout` to string -- that's what the `%s` formatting code does already. (I know the same pattern is used elsewhere in ElastiCluster's code, but it's a bug ;-))

elasticluster/providers/ec2_boto.py
```diff
@@ -206,18 +208,40 @@ def start_instance(self, key_name, public_key_path, private_key_path,
...
+                while  request.status.code != 'fulfilled':
+                    if timeout and time.time()-start_time > float(timeout):
+                        request.cancel()
+                        raise RuntimeError('spot instance timed out')
+                    time.sleep(10)
```
riccardomurri: Can you please make the `10` a class-level or module-level constant?

riccardomurri:
Merged, thank you very much for contributing!!

Issue #321, opened 2016-10-14
`elasticluster stop` terminates nodes sequentially


riccardomurri:
The code in `elasticluster stop` loops over the list of nodes and terminates them one by one.  This can take forever on large clusters.

We should adopt the same mechanism used in `elasticluster start` and use `multiprocessing.dummy` to parallelize cloud API invocations.

PR #323, opened 2016-10-15
Parallelizing elasticluster stop

cocoy:
Trying to test and fix feature #321


cocoy:
Re-sending with clean commit. closing for now

PR #324, opened 2016-10-15
Parallelizing stop.

cocoy:
Trying to fix parallelizing elasticluster stop #321

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -737,12 +737,19 @@ class Cluster(Struct):
         :param bool force:
           remove cluster from storage even if not all nodes could be stopped.
         """
+        failed = False
         log.debug("Stopping cluster `%s` ...", self.name)
+        nodes = self.get_all_nodes()
+        failed_nodes = self.stop_all_nodes_parallel(nodes,self.thread_pool_max_size,wait)
 
-        failed = self._stop_all_nodes(wait)
+        if (len(failed_nodes) > 0):
+           failed = True
 
         if failed:
             if force:
+                for n in  nodes:
+                    self.remove_node(n)
+
                 self._delete_saved_data()
                 log.warning(
                     "Not all cluster nodes have been terminated."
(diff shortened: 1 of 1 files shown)


elasticluster/cluster.py
```diff
@@ -763,38 +770,94 @@ def _delete_saved_data(self):
...
                     " assuming it has already been terminated.",
                     node.name, node.instance_id)
             except Exception as err:
-                failed += 1
                 log.error(
                     "Could not stop node `%s` (instance ID `%s`): %s %s",
-                    node.name, node.instance_id, err, err.__class__)
-        return failed
+                   node.name, node.instance_id, err, err.__class__)
+                is_Terminatd = False
```
riccardomurri: Typo: should be `is_Terminated` (there's a `e` missing`)

elasticluster/cluster.py
```diff
@@ -763,38 +770,94 @@ def _delete_saved_data(self):
...
+            exiting during node startup.
+            """
+            log.error(
+                "Interrupted: will save cluster state and exit"
+                " after all nodes have started.")
+            keep_running = False
+
+        # intercept Ctrl+C
+        with sighandler(signal.SIGINT, sigint_handler):
+            result = thread_pool.map_async(self._stop_node,node_stat_list)
```
riccardomurri: You can rewrite this so, and avoid creating two new lists::

```python
result = thread_pool.map_async((lambda node: self._stop_node(node, wait)), nodes)
```

riccardomurri: Or you can create a local function at line ~805::

```python
def _stop_node(node, wait=True):
  return self._stop_node(node, wait)
```

and then use it in the `map_async`::

```python
result = thread_pool.map_async(_start_node, nodes)
```

which looks to me the cleanest solution.

elasticluster/cluster.py
```diff
@@ -763,38 +770,94 @@ def _delete_saved_data(self):
...
+                    # instead to interrupt node creation as soon as possible?
+                    thread_pool.close()
+                    thread_pool.join()
+                    self.repository.save_or_update(self)
+                    # FIXME: should raise an exception instead!
+                    sys.exit(1)
+
+            # TODO: might return only failed node
+            return set(failed_node for failed_node, ok
+                        in itertools.izip(node_stat_list, result.get()) if not ok)
```
riccardomurri: If you go for the lambda/local function instead, `node_stat_list` here should be `nodes` instead.

elasticluster/cluster.py
```diff
@@ -763,38 +770,94 @@ def _delete_saved_data(self):
...
+    def _stop_node(node_wait):
+
+        """
+        node_wait= [ node, False]  is a list with those values
+        Stop the given node VM.
+        :return: bool -- True on success, False otherwise
+
+        Terminate the VM instance launched on the cloud for this specific node.
+        """
+        is_Terminated = False
```
riccardomurri: Please use PEP-8 conventions: `is_terminated` not `is_Terminated`.

riccardomurri:
LGTM, with a couple of exceptions:

1. There is a typo preventing the code from running.
2. We should also have a sequential-stop version (for debugging purposes) just like `Cluster.start()` does.  Can you please provide it too?

PR #326, opened 2016-10-17
Enable increase disk of the root volume  for ec2_boto provider

cocoy:
Sample parameter for increasing root disk.
``` 
[cluster/aws-slurm/frontend]
flavor=m1.small
boot_disk_size=20

[cluster/aws-slurm/compute]
flavor=m1.small
boot_disk_size=20
```

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -703,7 +703,9 @@ Optional configuration keys
     Define the size of boot disk to use.
     Only supported when the cloud provider is `google`.
     Values are specified in gigabytes.
-    Default value is 10.
+    Default value is 10 for `google`.
+    Now supports `ec2_boto` boot_disk_size. 
+    Default value is 8 for `ec2_boto`.
 
 ``tags``
     Comma-separated list of instance tags.
--- a/elasticluster/providers/ec2_boto.py
+++ b/elasticluster/providers/ec2_boto.py
@@ -256,7 +256,7 @@ class BotoCloudProvider(AbstractCloudProvider):
                     request=connection.get_all_spot_instance_requests(request_ids=request.id)[-1]
             else:
 
-                if not boot_disk_size or int( boot_disk_size > 0):
+                if not boot_disk_size or int( boot_disk_size > 8):
                     dev_sda1 = ec2.blockdevicemapping.EBSBlockDeviceType()
                     dev_sda1.size = int(boot_disk_size)
                     dev_sda1.delete_on_termination = True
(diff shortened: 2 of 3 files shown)


riccardomurri:
> Sample parameter for increasing root disk.

Thanks!  Will review next week, when I'm back from vacation.

cocoy:
Just notice that EBS volumes on AWS also has disk types, maybe we can add here similar to 'google' boot_disk_type parameter.            

### Disk Types
``` 
[ standard | gp2 | io1 ]
```

docs/configure.rst
```diff
@@ -703,7 +703,9 @@ Optional configuration keys
     Define the size of boot disk to use.
     Only supported when the cloud provider is `google`.
     Values are specified in gigabytes.
-    Default value is 10.
+    Default value is 10 for `google`.
+    Now supports `ec2_boto` boot_disk_size.
```
riccardomurri: "Now" will be in the past, 6 months from now :-)

Please use ElastiCluster version numbers instead, e.g. "Sets `boot_disk_size` on `ec2_boto` (starting ElastiCluster 1.3)"

elasticluster/providers/ec2_boto.py
```diff
@@ -214,10 +217,26 @@ def start_instance(self, key_name, public_key_path, private_key_path,
             security_groups = [security_group]
 
         try:
+
             #start spot instance if bid is specified
             if price:
                 log.info("Requesting spot instance with price `%s` ...", price)
-                request = connection.request_spot_instances(
+
+                if not boot_disk_size or int( boot_disk_size > 8):
```
riccardomurri: What did you intend this conditional to check, exactly?

- `not boot_disk_size` will only be true when `boot_disk_size == 0` (since it's supposed to be an integer)
- `int (boot_disk_size > 8)` will evalutate to 0 or 1 dependending on whether `boot_disk_size` is less or equal to 8
- `or`-ing them together will result in the condition being executed if `boot_disk_size==0` or if it's 9 or more.... is this what you really meant?

riccardomurri:
I am not sure the code as it stands currently will do what is intended...

elasticluster/providers/ec2_boto.py
cocoy: We have another AMI's (customized) that is bigger than the default Ubuntu and Amazon AMI 8GB sizes. Probably make it simplier comparing to default 8Gb rather than checking if boot_disk_size is set:
```
bdm = None
if boot_disk_type > 8:
   bdm = set to given disk_size  and disk
   
# spot instance
if price:
   .. 
   request = connection.request_spot_instances( ..
           block_device_map = bdm,
           ..

## ec2 default instances 
else:
    reservation = connection.run_instances (.. 
           block_device_map = bdm, 
```

Anyways, Closing this and cleaning this up put in separate git for the moment. Thanks for review @riccardomurri.

PR #338, opened 2016-11-08
Fixed a bug with a conditional which was always true.

hmartiniano:
OS version (for Ubuntu 14.04) wasn't being properly recognized causing ansible to try to restart  (non-existent) service the netfilter-persistent.
This commit fixes this problem (at least on Ubuntu 14.04).


riccardomurri:
Merging, thanks for your contribution!

PR #345, opened 2016-11-20
change pytest to tool:pytest to remove a warning in the tests

Moisan:
`py.test --cov=elasticluster` gives the following warning: 

`WC1 None [pytest] section in setup.cfg files is deprecated, use [tool:pytest] instead.`

This PR updates setup.cfg to remove that warning.


riccardomurri:
Merged, thanks for contributing!

PR #360, opened 2016-12-04
#183 / #267: Document configuration to get Ansible SSH pipelining to work on CentOS/RHEL

nicoulaj:
Related: #183 / #267

This documents a trick to get SSH pipelining to work on CentOS/RHEL: use the "User Data" script to configure sudo properly before Ansible connects.

(only tested on EC2)


riccardomurri:
Merged, thanks!

PR #361, opened 2016-12-04
SGE playbook: remove uses of ignore_errors

nicoulaj:
This replaces uses of `ignore_errors` with proper checks in the SGE playbook.
Also fixed a few "changed" flags.

### DIFF
--- a/elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
+++ b/elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
@@ -24,7 +24,8 @@
 
 - name: add the frontend as submission host
   action: shell bash -lc 'qconf -as {{groups.gridengine_master[0]}}'
-  ignore_errors: yes
+  register: command_result
+  changed_when: "'added to submit host list' in command_result.stdout"
   tags:
     - sge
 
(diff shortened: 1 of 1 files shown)


elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
```diff
@@ -35,27 +36,27 @@
...
-  action: shell bash -lc 'qconf -Ahgrp /root/allhosts.grp.conf'
-  ignore_errors: yes
+  action: shell bash -lc 'qconf -shgrpl | grep -q '@allhosts' || qconf -Ahgrp /root/allhosts.grp.conf'
+  register: command_result
+  changed_when: "'added \"@allhosts\" to host group list' in command_result.stdout"
   tags:
     - sge
 
 - name: Update @allhosts group
   action: shell bash -lc 'qconf -Mhgrp /root/allhosts.grp.conf'
```
riccardomurri: I'm unsure about this one, as it isn't protected by a conditional now - what would happen if you ran the playbook twice? (E.g., reconfigure a cluster that has already been successfully configured?)

elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
```diff
@@ -65,14 +66,16 @@
...
-  action: shell bash -lc 'qconf -Aq /root/all.q.conf'
-  ignore_errors: yes
+  action: shell bash -lc 'qconf -sq all.q &> /dev/null || qconf -Aq /root/all.q.conf'
+  register: command_result
+  changed_when: "'added' in command_result.stdout"
   tags:
     - sge
 
 - name: add hosts to the default queue
   action: shell bash -lc 'qconf -aattr queue hostlist @allhosts all.q'
```
riccardomurri: Same issue here, I think

elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
nicoulaj: That's the point of the change, when running the second time, no error will be shown in the logs.

In this case `qconf -shgrpl | grep -q '@allhosts'` will directly return 0 and `qconf -Ahgrp` will not be executed.

elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
nicoulaj: Same idea, if `all.q` already exists, `qconf -sq all.q` will return 0 and `qconf -Aq` will not be executed.

elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
riccardomurri: Looks like GitHub's display here is confusing :-)  My comment was about line 57: `qconf -Mhgrp` is executed unconditionally, as far as I can see.

elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
nicoulaj: Yes, this call is idempotent:
```
$ qconf -aattr queue hostlist @allhosts all.q
No modification because "@allhosts" already exists in "hostlist" of "all.q"
$ echo $?
0
```

elasticluster/share/playbooks/roles/gridengine/tasks/master.yml
nicoulaj: Oh, sorry, I got confused :)
This task was already idempotent from my tests:
```
$ qconf -Mhgrp /root/allhosts.grp.conf
root@ip-172-31-9-177.eu-west-1.compute.internal modified "@allhosts" in host group list
$ echo $?
0
```

riccardomurri:
Merged, thanks a lot for contributing!

PR #362, opened 2016-12-06
Put short hostnames first in hosts file

nicoulaj:
With current configuration, SGE uses the `ip-XXX-XXX-XXX-XXX` in `qhost`, and `ip-XXX-XXX-XXX-XXX.eu-west-1.compute.internal` in `qstat`, which makes pretty hard to administer the cluster...

SGE picks the name in first position in `/etc/hosts` ([related mailing list thread](http://gridengine.org/pipermail/users/2016-December/009469.html)).

Only tested on EC2 with CentOS 6/7 and SGE+Ganglia playbooks.

### DIFF
--- a/elasticluster/share/playbooks/roles/common/templates/etc/hosts.j2
+++ b/elasticluster/share/playbooks/roles/common/templates/etc/hosts.j2
@@ -7,5 +7,6 @@
 
 {% for host in hosts|default(groups['all'])|sort if not host.startswith('localhost') %}
 {%- set fqdn = hostvars[host].ansible_fqdn -%}
-{{hostvars[host].ansible_default_ipv4.address}} {% if not fqdn.startswith('localhost') %}{{fqdn}} {{fqdn.split('.')[0]}}{% endif %} {{host}}
+{{hostvars[host].ansible_default_ipv4.address}} {{host}} {% if not fqdn.startswith('localhost') %}{{fqdn}} {{fqdn.split('.')[0]}}{% endif %}
+
 {% endfor -%}


riccardomurri:
Hmmm... the [hosts(5)][a] man page states (emphasis mine):

> For each host a single line should be present with the following information:
> 
>        IP_address canonical_hostname [aliases...]
> 
> [...]
> Optional aliases provide for name changes, alternate spellings, **shorter
> hostnames,** or generic hostnames (for example, localhost).

However, I agree with you that the change you propose in this PR will probably
not break anything and just make life simpler for GridEngine admins.

Let me think a bit about it.

[a]: http://manpages.ubuntu.com/manpages/xenial/en/man5/hosts.5.html

riccardomurri:
Found [this comment][1] in a Linux support forum:

> if you don't list the FQDN hostname as first itme, some Linux/BSD
> distributions won't return FQDN hostname with command 'hostname -f'. [...] So,
> if 'hostname -f' returns correct FQDN hostname on your server, order doesn't
> matter at all.

Can you please check whether this is the case?

[1]: http://www.iredmail.org/forum/post29687.html#p29687

riccardomurri:
A more thorough discussion: http://unix.stackexchange.com/questions/77917/include-domain-name-or-not-in-etc-hosts

nicoulaj:
Well this [answer](http://unix.stackexchange.com/a/255889/2231) is pretty scary...

`hostname -f` shows before patch: `ip-XXX-XXX-XXX-XXX.eu-west-1.compute.internal`, after patch: `master001`.

I doubt it will break anything, but maybe the wise decision here is not to merge this as I cannot test all platforms/playbooks... Or maybe make it optional via Ansible tags...

riccardomurri:
Scary, but also quite vague and undocumented.  I'd ignore it.

So far, the only demonstrable effect is that `hostname -f` would return a non-fully-qualified hostname. However, the 2nd comment to [this answer](http://unix.stackexchange.com/a/77924/274) states that `libnss-myhostname` can provide the correct FQDN independently of entries in `/etc/hosts`.  Now, `libnss-myhostname` is not available across all distributions:

* in Ubuntu, [since (at least) 12.04](http://packages.ubuntu.com/search?keywords=libnss-myhostname&searchon=names&suite=all&section=all)
* In Debian, [since Debian 7 "wheezy"](https://packages.debian.org/search?keywords=libnss-myhostname&searchon=names&suite=all&section=all)
* CentOS seems not to provide it, neither in version 6.x nor 7.x (version 7 ships the man page in the "systemd" package but the library is not there)

I am inclined to merge, but I would like to provide `nss-myhostname` as a "safety net"...

riccardomurri:
I stand corrected: RHEL/CentOS 7.x ships `/lib64/libnss_myhostname.so.2` in the `systemd-libs` package. So the only distro left out is CentOS 6.x

riccardomurri:
Seems to be an issue with SLURM clusters as well after reboot, let's merge and see what happens.

riccardomurri:
Merged, thanks!

PR #378, opened 2017-01-23
Fixed gce.py so that ~ can be used in user_key_* parameters of config.

ikhaja:
Made fix so that relative paths using ~/ can be used with parameters user_key_private and user_key_public in the login section of .elasticluster/config for GCE.

Executing the command `elasticluster start my_cluster` with the following login section:
```
[login/google]
image_user=my_username
image_user_sudo=root
image_sudo=True
user_key_name=elasticluster
user_key_private=~/.ssh/google_compute_engine
user_key_public=~/.ssh/google_compute_engine.pub
```
Produces errors such as the following:
```
2017-01-23 06:35:33 ncbiblast-2 gc3.elasticluster[16705] ERROR Could not start node `frontend001`: [Errno 2] No such file or directory: '~/.ssh/google_compute_engine.pub' -- <type 'exceptions.IOError'>
```

### DIFF
--- a/elasticluster/providers/gce.py
+++ b/elasticluster/providers/gce.py
@@ -307,7 +307,13 @@ class GoogleCloudProvider(AbstractCloudProvider):
         else:
             instance_id = 'elasticluster-%s' % uuid.uuid4()
 
-        public_key_content = file(public_key_path).read()
+        # Use the with keyword so that the file is properly closed.
+        # For os.path.expanduser, if the expansion fails or if the path does
+        # not begin with a tilde, the path is returned unchanged. If the
+        # returned path is a relative path, determine the absolute path and use
+        # that to open the file.
+        with open(os.path.abspath(os.path.expanduser(public_key_path)), 'r') as f:
+            public_key_content = f.read()
 
         instance = {
             'name': instance_id,


riccardomurri:
Hello @ikhaja thanks for contributing!

What version of the ElastiCluster code are you running?  I think this is already fixed in the current "master" branch, although through a different route: pass the `user_*_key` attributes through a validation function (https://github.com/gc3-uzh-ch/elasticluster/blob/master/elasticluster/conf.py#L124) which ultimately calls `file_name` which expands `~/` references (https://github.com/gc3-uzh-ch/elasticluster/blob/master/elasticluster/validate.py#L46)

ikhaja:
Hello Riccardo!

Thank you for your reply.  I am running the git clone of the current master branch for ElastiCluster on Google Cloud using Ubuntu 16.04.1 LTS virtual machine.

To me it seems as though in providers/gce.py on line 310: `public_key_content = file(public_key_path).read()` does not get a value when ~ is in the path.

Please let me know if I can be of more help.

Ihtisham

riccardomurri:
Hi @ikhaja, I have added a test that `os.path.expanduser` is called when reading the `[login/*]` section in commit b37e36e.

The test passes already with the current code, so I'd say that the feature is already supported so I'm closing this PR.  Let me know if you think the test is not correctly covering the feature you wanted and re-open the PR.

I've merged in your `with ... = f.read()` suggestion in the commit following the one above.

Thank you very much for your contribution!

riccardomurri:
I stand corrected: `os.expanduser` is called but somehow its return value does not make it to the `Cluster` instance... so, yes, you're right.

riccardomurri:
Ok, so here's the story: it was supposed to be handled already by the current code, but I did not invoke the validation functions correctly, so the converted values were thrown away and the original ones were being used instead... It was a bug, which has now been fixed.

Thank you very much for raising the issue!

ikhaja:
Hello Riccardo!
Thank you greatly for making the fixes! Sorry I have been inattentive for the last couple of days. I will try it out very soon and will let you how it works in my GCE environment :)

\*****
Ihtisham Khaja

ikhaja:
Hello Riccardo!
Just confirming that it is working great on GCE. Again, thank you!

\*****
IK

PR #383, opened 2017-01-24
Change mentions of slurm_workers to slurm_worker in the documentation

Moisan:
While the code has been changed to use `slurm_worker` everywhere and warn the user if `slurm_workers` was used, the documentation has not been changed yet.


riccardomurri:

PR #410, opened 2017-03-13
Pmess

witlox:
A lot of changes regarding mesos, spark, security groups fix and kafka


elasticluster/share/playbooks/site.yml
```diff
@@ -32,6 +34,7 @@
 # Jupyter installation comes last, to allow it to pick up any SW that's been
 # installed so far (e.g., for kernels, or for ipyparallel)
 - include: roles/jupyterhub.yml
+- include: roles/rstudio.yml
```
riccardomurri: AFAICS, this just runs one role that installs a single SW... So far, I've kept independent single-role plays as an option in `after.yml`. IOW, no cluster no playbook.  (This can of course be changed, but I'm a bit afraid of an explosion in the number of playbooks.)

elasticluster/share/playbooks/roles/rstudio/templates/etc/rstudio/rsession.conf.j2
```diff
@@ -0,0 +1,3 @@
+# {{ ansible_managed }}
```
riccardomurri: Please use the "MANAGED BY ELASTICLUSTER" header (see, e.g., `common/templates/etc/hosts.j2`)

elasticluster/share/playbooks/roles/rstudio/templates/etc/rstudio/rserver.conf.j2
```diff
@@ -0,0 +1,9 @@
+# {{ ansible_managed }}
```
riccardomurri: Please use the "MANAGED BY ELASTICLUSTER" header (see, e.g., `common/templates/etc/hosts.j2`)

elasticluster/share/playbooks/roles/rstudio/templates/etc/rstudio/rsession.conf.j2
```diff
@@ -0,0 +1,3 @@
...
+
+# R Session Configuration File
```
riccardomurri: This is basically an empty file. Why do we need it?

elasticluster/share/playbooks/roles/rstudio/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+    deb: "r-studio-server/{{ rstudio_server_version }}_amd64.deb"
+
+
+- name: update server configuration file
+  template:
+    src: etc/rstudio/rserver.conf.j2
+    dest: /etc/rstudio/rserver.conf
+    owner: root
+    group: root
+    mode: 0644
```
riccardomurri: Files managed by ElastiCluster should have mode `0444` to discourage local editing.

elasticluster/share/playbooks/roles/rstudio/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
+---
+
+
+- name: Check playbook compatibility with host platform
+  fail:
+    msg="R Studio Server installation only works on Debian/Ubuntu presently."
+  when: not is_debian_or_ubuntu and not {{experimental}}
+
+
+- name: create download directory
```
riccardomurri: Other roles just use `/var/tmp` (or allow to set a location, defaulting to `/var/tmp`)

elasticluster/share/playbooks/roles/rstudio.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: R-Studio playbook
+  hosts: rstudio
+  roles:
+    - rstudio
+  tags:
+    - rstudio
```
riccardomurri: maybe also add `r` as a tag?

elasticluster/share/playbooks/roles/mesos/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+mesos_install_mode: "master" # {master|slave}
+
+quorum_size: "{{groups['mesos_master']|length}}"
```
riccardomurri: If I understand this correctly, the `quorum_size` is only used as an auxiliary quantity in the computation of `mesom_quorum`.  If this is correct, I would rather use Jinja's `{% set %}` directive, rather than creating a role-level variable. This also avoids having to convert again from string to int::

    mesos_quorum: "{% set nhosts = groups.mesos_master|length|int %}{if nhosts < 3 %}1{% else %}{{ (nhosts * 2 / 3)|round|int }}{% endif %}"

elasticluster/share/playbooks/roles/mesos.yml
```diff
@@ -0,0 +1,29 @@
+---
+
+- name: Install Mesos master node
+  hosts: mesos_master
+  vars:
+    java_version: 8
+    zookeeper_hosts: "mesos_master"
```
riccardomurri: In other playbook, I have tied variables to the roles they refer to (better explicit than implicit)::

    roles:
      - role: zookeeper
        zookeeper_hosts: "mesos_master"
      # ...

riccardomurri: I find the variable name `zookeepr_hosts` quite misleading.  This is the name of Ansible *group*, whereas from the var name I would expect a list of hosts...

elasticluster/share/playbooks/roles/kafka/templates/server.properties.j2
```diff
@@ -0,0 +1,99 @@
+############################# Server Basics #############################
+
+{% for server in groups['stream_master'] %}
+{% if server == inventory_hostname %}
+broker.id={{ server[-3:] | int }}
```
riccardomurri: No, no, no -- see comment in `zookeeper/templates/zoo.cfg.j2` line 12: use `{{ loop.index }}`

elasticluster/share/playbooks/roles/kafka/templates/server.properties.j2
```diff
@@ -0,0 +1,99 @@
...
+port={{ kafka_port + (server[-3:] | int) }}
```
riccardomurri: Same here: use `{{ loop.index }}`

elasticluster/share/playbooks/roles/zookeeper/templates/myid.j2
```diff
@@ -0,0 +1,5 @@
+{% for server in groups[zookeeper_hosts] %}
```
riccardomurri: This use of `zookeeper_hosts` forces it to be an Ansible group, i.e., exposes an internal detail up to the user config level -- can't we just have `zookeeper_hosts` be set to a list of hosts?  (Defaulting to `groups.zookeeper_hosts|default([])`)

elasticluster/share/playbooks/roles/zookeeper/defaults/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+
+zookeeper_hosts: "zookeeper"
+
+apt_cache_timeout: 3600
```
riccardomurri: Doesn't belong in here.

elasticluster/share/playbooks/roles/zookeeper/defaults/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+
+zookeeper_hosts: "zookeeper"
```
riccardomurri: See comment on line 1 of `myid.j2` below.

elasticluster/share/playbooks/roles/mesos/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+- name: Install Mesosphere on master nodes (Redhat/Centos)
+  yum: name={{ item }} state=present
+  with_items:
+    - mesosphere
+  when: not is_debian_or_ubuntu and mesos_install_mode == "master"
+
+- name: Mesos ZK config file
+  template: src=zk.j2 dest=/etc/mesos/zk
+
+- name: Mesos master quorum file
```
riccardomurri: I would move all "master" actions into a file `master.yml` and then use conditional inclusion:

```
- include: master.yml
  when: mesos_install_mode == "master"
```

elasticluster/share/playbooks/roles/mesos/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- name: Mesos master IP file
+  template: src=master-ip.j2 dest=/etc/mesos-master/ip
+  when: mesos_install_mode == "master"
+
+- name: Mesos master hostname file
+  template: src=master-hostname.j2 dest=/etc/mesos-master/hostname
+  when: mesos_install_mode == "master"
+
+- name: Mesos slave IP file
```
riccardomurri: I would move all "slave" actions into a separate file `slave.yml` and then use conditional inclusion:

```
- include: slave.yml
  when: mesos_install_mode == "slave"
```

elasticluster/share/playbooks/roles/mesos.yml
```diff
@@ -0,0 +1,29 @@
...
+  tags:
+    - zookeeper
+    - mesos
+    - marathon
+    - master
+  roles:
+    - zookeeper
+    - {role: mesos, mesos_install_mode: "master" }
```
riccardomurri: The way similar roles have been handled so far is to have separate roles `mesos-master`, `mesos-slave` and (possibly) `mesos-common`.

I understand this is a matter of style, though, and the commonalities in this case may be like 80% of the code, so I won't fuss on it :-)

elasticluster/share/playbooks/roles/kafka/defaults/main.yml
```diff
@@ -0,0 +1,28 @@
+---
```
riccardomurri: Do we really need all these customization points for Kafka?

elasticluster/share/playbooks/roles/kafka/defaults/main.yml
```diff
@@ -0,0 +1,28 @@
...
+
+kafka_log_dir: /var/log/kafka
+kafka_log_flush_interval_messages: 10000
+kafka_log_flush_interval_ms: 1000
+kafka_log_retention_bytes: 104857600  #100 M
+kafka_log_retention_hours: 24
+kafka_log_segment_bytes: "{{kafka_log_retention_bytes}}"
+kafka_log_cleanup_interval_mins: 1
+
+monasca_wait_for_period: 30
```
riccardomurri: These should be prefixed with `kafka_...`

elasticluster/share/playbooks/roles/hadoop-common/defaults/main.yml
```diff
@@ -5,4 +5,4 @@
 # symlinks `/etc/hadoop/conf` to the actual configuration
 HADOOP_CONF_DIR: '/etc/hadoop/conf.elasticluster'
 
-YARN_NM_LOG_DIR: '/yarn/apps'
\ No newline at end of file
+REPLICATION_FACTOR: 3
```
riccardomurri: Please prefix variable names with the role they refer to; this should be called e.g. `HADOOP_HDFS_REPLICATION_FACTOR`

Ansible has a single namespace for variables, so if some other role defines a `replication_factor` variable, the net result will be strictly dependent on the order roles are applied, which can lead to endless fun in debugging...

riccardomurri: (Also: the default here was 1, which is correct for Science Cloud and -I presume- for most IaaS cloud systems with highly-available storage.  Is there a reason for the change?)

elasticluster/share/playbooks/roles/bigtop/README.rst
```diff
@@ -0,0 +1,35 @@
...
+=========
+
+Install BigTop_ and its dependencies to provide the binaries for the Apache Hadoop Component Stack.
+
+Role Variables
+--------------
+
+The following variable may be set to alter the role behavior:
+
+``EXPERIMENTAL``
```
riccardomurri: Please prefix variable names with the role they refer to; this should be called e.g. `BIGTOP_EXPERIMENTAL`

Ansible has a single namespace for variables, so if some other role defines a `experimental` variable, the net result will be strictly dependent on the order roles are applied, which can lead to endless fun in debugging...

elasticluster/share/playbooks/roles/java/tasks/main.yml
```diff
@@ -2,29 +2,24 @@
...
-  when: is_debian_or_ubuntu and "java_version == 8"
-
-
-- name: Installing openjdk-doc (Debian/Ubuntu)
-  apt: name=openjdk-{{ java_version }}-doc state=present update_cache=yes cache_valid_time=8600
-  when: is_debian_or_ubuntu
+  apt_repository:
+    repo: ppa:openjdk-r/ppa
+    state: present
+  when: is_debian_compatible and "java_version == 8"
```
riccardomurri: Why the quotes around `"java_version == 8"`?  Shouldn't it be `java_version|int == 8`?

My understanding of Ansible/Jinja2 conditionals (poor, I'll admit) is that it will not be parsed as a conditional expression, rather as a non-empty string, which *always* evaluates to true...  IOW, jave you tried running it with `java_version` set to 7?

elasticluster/share/playbooks/roles/jupyter/defaults/main.yml
```diff
@@ -14,4 +14,4 @@ PYTHON3_EXE: '/opt/anaconda3/bin/python3'
 R_EXE: '/usr/bin/R'
 
 # the Python interpreter used to rune Jupyter/JupyterHub
-jupyter_python: '{{PYTHON3_EXE}}'
\ No newline at end of file
+jupyter_python: '{{PYTHON2_EXE}}'
```
riccardomurri: Why this change? Installation instructions for Jupyter recommend running it with the Py3 interpreter.

elasticluster/share/playbooks/roles/rstudio/templates/etc/rstudio/rserver.conf.j2
```diff
@@ -1,5 +1,3 @@
-# {{ ansible_managed }}
-
```
riccardomurri: Please *do* use a header indicating that the file is managed by ElastiCluster; see an example in `playbooks/roles/slurm-common/templates/slurm.conf.j2`:

```
# THIS FILE IS CONTROLLED BY ELASTICLUSTER
# local modifications will be overwritten
# the next time `elasticluster setup` is run!
#
```

elasticluster/share/playbooks/roles/spark-common/defaults/main.yml
```diff
@@ -5,4 +5,5 @@
 # symlinks `/etc/spark/conf` to the actual configuration
 SPARK_CONF_DIR: '/etc/spark/conf.elasticluster'
 
-master: True
+history_dir: '/user/spark/applicationHistory'
```
riccardomurri: Please prefix variables names... (see above comments)

riccardomurri:
By the way: BigTop has released 1.2.0 in the meanwhile.  Do we still need the "experimental" stuff and our own repository?

elasticluster/share/playbooks/roles/zookeeper/meta/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+dependencies:
+  - { role: java }
```
witlox: not necessarily, zookeeper can be used independent of hadoop components

riccardomurri: ...and where are pulling ZK packages from then?

witlox: for debian-esque it's in the universe repo, for redhat-esque it's in the standard repo

witlox: just an older version than packaged with bigtop, for example mesos-master uses the distribution defaults

riccardomurri: hmmm... that means that the role installs and does different things, depending on whether you install it in conjunction with Spark/Hadoop or not...  I would prefer a more consistent/predictable behavior, as it makes debugging easier.

witlox: sounds reasonable, shall I create a zookeeper play for hadoop, and one for mesos?

riccardomurri: why not just add bigtop as a dependency and always install from there?

riccardomurri:
I have spent some time rebasing this on top of current "master" so it's clearer what's changed and what not.
Can you please switch to the rebased commits?  It's all on https://github.com/riccardomurri/elasticluster on branch "feature/bigtop-1.2.0-et-al"

elasticluster/share/playbooks/roles/jupyter/files/usr/local/share/jupyter/kernels/pyspark2/startup.py
```diff
@@ -23,12 +23,9 @@
 
 # make `pyspark` and `py4j` available to this process
 spark_home_python = join(spark_home, "python")
-sys.path.insert(0, spark_home_python)
-sys.path.insert(0, join(spark_home_python, 'pyspark'))
-sys.path.insert(0, join(spark_home_python, 'lib/py4j-0.8.2.1-src.zip'))
```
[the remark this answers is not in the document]
danielspicar: Currently the hardcoded version of py4j-0.8.2.1 does not match the version that gets installed. As a result the kernel does not work as expected. At least this should be changed to lib/py4j-src.zip which is a link to the currently installed version (0.10.4).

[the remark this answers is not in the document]
riccardomurri: @danielspicar Note that support for Spark 2.1.1 (via Apache Bigtop 1.2.1) has already been merged in ElastiCluster's "master" branch. This part of the pull request is outdated.

PR #412, opened 2017-03-28
Fixing 100% CPU on gmond for cluster nodes

akmb:
After starting a cluster (slurm, ganglia) I discover that gmond uses 100% on all compute nodes. This happens on ubuntu 14.04 and 16.04.
This fix differ configuration between ganglia_monitor and ganglia_master.


riccardomurri:
Merged in commit  e721d9d -- thanks a lot for contributing!

(Note the commit hash is different from the one on this PR, as I've done some minor edits.)

PR #413, opened 2017-03-30
Added a LICENCE file for GPLv3 to the base of the project.

lila:
added a license file to the top level project with the contents of the GPLv3 from https://www.gnu.org/licenses/gpl-3.0.txt .  Having this file indicating the license will allow me to submit patches to the project.


riccardomurri:
Thanks a lot, not having a licence file was quite an oversight!!!

Issue #381, opened 2017-01-24
 boot_disk_size in the cluster node section is ignored on GCE


marcbrisson:
Using google compute engine. I the boot_disk_size in the cluster section is ignored.

[cluster/slurm]
cloud=google
login=google
setup=slurm
security_group=default
image_id=ubuntu-1604-xenial-v20170113
flavor=n1-standard-1
master_nodes=1
worker_nodes=4
ssh_to=master
image_userdata=
boot_disk_size=15

all machines are 10gb only, the default size on GCE

riccardomurri:
@marcbrisson can you please check that @Moisan 's pull request #419 fixes the issue?

marcbrisson:
Yes it looks good!

riccardomurri:
Thanks for taking the time to check!

The issue has thus just been fixed by merging #419 , contributed by @Moisan

PR #419, opened 2017-04-06
Add boot_disk_size, boot_disk_type, scheduling, node_name and tags to…

Moisan:
… the nodes section for gce

This should fix issue #381. The issue was coming from the fact that cloud specific parameters in the config files weren't passed to the Node instances.

### DIFF
--- a/elasticluster/conf.py
+++ b/elasticluster/conf.py
@@ -599,6 +599,11 @@ def _gather_node_kind_info(kind_name, cluster_name, cluster_conf):
             'login',
             'network_ids',
             'security_group',
+            'node_name',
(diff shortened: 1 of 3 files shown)


riccardomurri:
Looks good to me, thanks!

@marcbrisson can you please confirm that this fixes #381 for you?

Moisan:
Travis fails on a test that made sure boot_disk_size wasn't in the Node extra. To get the boot_disk_size to work it has to be in the extra attribute. I'll add a commit to fix the test.

riccardomurri:
Thanks a lot for taking the time to dig into this and contribute!

PR #428, opened 2017-05-08
Initial Keystone v3 support

cbrown2:
The implements keystone v3 support but I'm not clear if it breaks support for keystone v2 only environments. I suspect it may as keystone v2 cannot handle the two new variables. Unfortunately I have no v2 only environments to test this on. Perhaps this can be used as a starting point?

Alternative would perhaps be to implement another openstack provider for keystone v3 (a bit nasty) or detect API version in the current openstack provider, probably using environmental variable OS_IDENTITY_API_VERSION

### DIFF
--- a/elasticluster/conf.py
+++ b/elasticluster/conf.py
@@ -180,6 +180,8 @@ CLOUD_PROVIDER_SCHEMAS = {
         Optional("auth_url", default=os.getenv('OS_AUTH_URL', '')): url,
         Optional("username", default=os.getenv('OS_USERNAME', '')): nonempty_str,
         Optional("password", default=os.getenv('OS_PASSWORD', '')): nonempty_str,
+        Optional("user_domain_id", default=os.getenv('OS_USER_DOMAIN_NAME', '')): nonempty_str,
+        Optional("project_domain_id", default=os.getenv('OS_PROJECT_DOMAIN_NAME', '')): nonempty_str,
         Optional("project_name",
                  # if OS_PROJECT_NAME is not defined,
                  # try legacy variable OS_TENANT_NAME as a fallback
--- a/elasticluster/conf.py
+++ b/elasticluster/conf.py
@@ -180,8 +180,8 @@ CLOUD_PROVIDER_SCHEMAS = {
         Optional("auth_url", default=os.getenv('OS_AUTH_URL', '')): url,
         Optional("username", default=os.getenv('OS_USERNAME', '')): nonempty_str,
         Optional("password", default=os.getenv('OS_PASSWORD', '')): nonempty_str,
-        Optional("user_domain_id", default=os.getenv('OS_USER_DOMAIN_NAME', '')): nonempty_str,
-        Optional("project_domain_id", default=os.getenv('OS_PROJECT_DOMAIN_NAME', '')): nonempty_str,
+        Optional("user_domain_name", default=os.getenv('OS_USER_DOMAIN_NAME', '')): nonempty_str,
+        Optional("project_domain_name", default=os.getenv('OS_PROJECT_DOMAIN_NAME', '')): nonempty_str,
         Optional("project_name",
                  # if OS_PROJECT_NAME is not defined,
                  # try legacy variable OS_TENANT_NAME as a fallback
--- a/elasticluster/providers/openstack.py
+++ b/elasticluster/providers/openstack.py
@@ -64,8 +64,8 @@ class OpenStackCloudProvider(AbstractCloudProvider):
     :param str username: username of the keystone user
     :param str password: password of the keystone user
     :param str project_name: name of the project to use
-    :param str user_domain_id: name of the user domain to use
-    :param str project_domain_id: name of the project domain to use
+    :param str user_domain_name: name of the user domain to use
+    :param str project_domain_name: name of the project domain to use
     :param str auth_url: url of keystone endpoint
     :param str region: OpenStack region to use
     :param str storage_path: path to store temporary data
@@ -77,7 +77,7 @@ class OpenStackCloudProvider(AbstractCloudProvider):
     __node_start_lock = threading.Lock()  # lock used for node startup
 
     def __init__(self, username, password, project_name, auth_url,
-                 user_domain_id, project_domain_id,
+                 user_domain_name, project_domain_name,
                  region_name=None, storage_path=None,
                  request_floating_ip=False,
                  nova_api_version=DEFAULT_OS_NOVA_API_VERSION):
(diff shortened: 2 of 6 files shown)


riccardomurri:
Hi @cbrown2 -- thanks a lot for working on this! I'm so sorry for the late reply to this PR, too many tasks to juggle...

We run a dual v2/v3 keystone stack here so I can try and test it, when you tell me it's ready (I notice you're still committing / updating).

riccardomurri:
Re: detecting v2/v3 -- maybe the code can take the decision based on the presence of the `OS_PROJECT_DOMAIN` env var or configuration setting?  Is it required for Keystone v3?

riccardomurri:
hi @arcimboldo any opinions on how to handle the v2/v3 issue? how should the code detect which version of the Keystone API to use?

verdurin:
@riccardomurri just so you know I've been testing this and there are still some problems with it. Notably, it still has the `pbr` error that you've been working on lately.

riccardomurri:
@cbrown2 I have submitted a PR to your repo merging changes from the latest "master".  You should be able to merge it and rebase on "master" w/out conflicts.

elasticluster/share/etc/config.template
```diff
@@ -132,6 +132,14 @@
 #                    command. If an environment variable
 #                    `OS_TENANT_NAME` is set, this option is ignored.
 #
+# user_domain_id:    OpenStack user domain to use,
```
arcimboldo: id != name, please make it consistent.

elasticluster/share/etc/config.template
```diff
@@ -132,6 +132,14 @@
...
+#                    If an environment variable
+#                    `OS_USER_DOMAIN_NAME` is set, this option is ignored.
+#
+# project_domain_id: OpenStack project domain to use,
```
arcimboldo: Same as before: domain_id != domain_name, be consistent

elasticluster/providers/openstack.py
```diff
@@ -506,9 +518,15 @@ def __setstate__(self, state):
         self._os_username = state['username']
         self._os_password = state['password']
         self._os_tenant_name = state['project_name']
+        self._os_user_domain_name = state['user_domain_name']
+        self._os_project_domain_name = state['project)domain_name']
```
arcimboldo: typo: `project)domain_name` => `project_domain_name`

elasticluster/share/etc/config.template
```diff
@@ -181,8 +189,11 @@ ec2_region=nova
 [cloud/hobbes]
 provider=openstack
 auth_url=http://hobbes.gc3.uzh.ch:5000/v2.0
+# domain_name=Default
 # username=***REPLACE WITH YOUR USERNAME (OS_USERNAME)***
 # password=***REPLACE WITH YOUR PASSWORD (OS_PASSWORD)***
+# project_domain_name=***REPLACE WITH YOUR PROJECT DOMAIN (OS_PROJECT_DOMAIN_NAME)***
+# project_user_name=***REPLACE WITH YOUR USER DOMAIN (OS_PROJECT_USER_NAME)***
```
arcimboldo: what's `project_user_name`? AFAIK it's never used.

tests/_helpers/config.py
```diff
@@ -53,6 +53,8 @@
 auth_url = http://openstack.example.com:5000/v2.0
 username = ${USER}
 password = XXXXXX
+project_domain_id = Default
```
arcimboldo: As before: domain_id != domain_name, be consistent

tests/_helpers/config.py
```diff
@@ -53,6 +53,8 @@
...
+user_domain_id = Default
```
arcimboldo: btw, this is the *name* of the default domain in openstack, while the corresponding *id* is (quite confusing) `default`

tests/_helpers/config.py
```diff
@@ -360,9 +364,11 @@ def get_config(self, path):
                     },
                 "cloud": {
                     "provider": "openstack",
-                    "auth_url": "http://cloud.gc3.uzh.ch:5000/v2.0",
+                    "auth_url": "http://cloud.gc3.uzh.ch:35357",
                     "username": "myusername",
                     "password": "mypassword",
+                    "project_domain_id": "Default",
```
arcimboldo: Again domain_id != domain_name and `Default` is the name of default domain with id `default` so this is wrong anyway.

elasticluster/conf.py
```diff
@@ -180,6 +180,8 @@
         Optional("auth_url", default=os.getenv('OS_AUTH_URL', '')): url,
         Optional("username", default=os.getenv('OS_USERNAME', '')): nonempty_str,
         Optional("password", default=os.getenv('OS_PASSWORD', '')): nonempty_str,
+        Optional("user_domain_name", default=os.getenv('OS_USER_DOMAIN_NAME', '')): nonempty_str,
```
arcimboldo: Note: in some cases it's possible that the environment contains OS_USER_DOMAIN_ID instead of OS_USER_DOMAIN_NAME, it would be nice to:
* read the _ID variable if _NAME is not present
* create a new instance of v3.Password using the ID instead of the name
* possibly suggest the user to put the name instead of the id in the configuration file.

elasticluster/providers/openstack.py
```diff
@@ -113,13 +118,16 @@ class OpenStackCloudProvider(AbstractCloudProvider):
     __node_start_lock = threading.Lock()  # lock used for node startup
 
     def __init__(self, username, password, project_name, auth_url,
```
arcimboldo: Better would be to be able to also use the *domain_id arguments if possible.

riccardomurri:
@cbrown2 I have pushed further changes into my update to this PR, that make it work with both v2 and v3 (select via config option, or try both and use first that works).  Will you be able to apply the further edits suggested by @arcimboldo ?  After that, I think we're good to go.

@verdurin Will you have time to test?  Branch "feature/keystone-api-v3" at http://github.com/riccardomurri/elasticluster should work now with both v2 and v3

cbrown2:
@riccardomurri no problem regarding delay, this code was really just the domain parameters plus some additional python imports to prove that I could make it work with v3. Lovely to see some changes which allow it to work with both v2 AND v3

@arcimboldo thanks for the review, I'm currently on holiday so won't have time to look at these until next week. Feel free to make the requested changes yourself if you wish and while it is still fresh in the mind?

cbrown2:
With regards to detecting the version, I thought about using OS_IDENTITY_API_VERSION env variable? I'm not sure endpoints will actually be set as /v3 in future as my understanding is there is no requirement to actually tag /v3 to the end for keystone v3 API onwards. But I may be completely wrong!

verdurin:
@riccardomurri testing now. There are some requirements on the cloud I'm testing from that are missing, so I can't test fully, but I was able to build and run, and I saw an error message about missing keystone `V2` and `V3` access, before I had configured the `auth_url`, so it's very encouraging.

verdurin:
Okay, this seems to work in very basic testing, as long as both `project_domain_name` and `user_domain_name` are set. 

Would be helpful to indicate in the docs that these two are mandatory for keystone v3.

cbrown2:
@riccardomurri hi, I'm back from holiday. What do you need, if anything, from me? If you are just waiting on the edits, would you or @arcimboldo mind applying them as I've kind of lost track of what needs doing where. I seem to have a PR against my fork, then theres a feature branch on another fork.... slughtly confused.

riccardomurri:
Hi Christopher,

Now it's my turn to be on vacation 😀 I'll be back to work on Monday.

IIRC, the feature branch on my fork is the latest and mostly correct
working code. I'd say we would need (in priority order):

1. Testing that the code on the feature branch works both with Keystone API
v2 and v3.

2. Using environment variable OS_IDENTITY_API_VERSION to choose between v2
and v3 and rename the config item accordingly (I have used
"keystone_api_version").

3. Implement the suggestions by @arcimboldo about name/I'd.

I personally would already merge after 1. and 2. , and leave 3. as an
enhancement.

riccardomurri:
Squashed and merged (with some additions and reworkings) in commit 50a958d and following.

Thanks a lot for contributing!  This is a very welcome addition and a badly needed feature :-)

PR #429, opened 2017-05-10
added libcloud

witlox:


docs/configure.rst
```diff
@@ -302,6 +302,18 @@ Valid configuration keys for *openstack*
     instance doesn't get one automatically.
 
 
+Valid configuration keys for *libcloud*
+----------------------------------------
+
+    See: https://libcloud.readthedocs.io/en/latest/supported_providers.html
+    for all available providers. Configuration keys are provider dependent.
```
riccardomurri: I wouldn't know how to write a configuration stanza for the `libcloud` driver based on these terse instructions: libcloud does not provide a configuration file format, only a Python API. How are the configuration keys in the ElastiCluster mapped to libcloud's Python API?   At least a short list of mandatory/minimal/common keys should be given.

docs/configure.rst
```diff
@@ -302,6 +302,18 @@ Valid configuration keys for *openstack*
...
+
+
+``driver_name``:
+
+    Name of the driver you want to connect with.
```
riccardomurri: The "driver" has been called a "provider" in the previous paragraph...

elasticluster/cluster.py
```diff
@@ -583,7 +579,7 @@ def _compute_min_nodes(self, min_nodes=None):
         if min_nodes is None:
             min_nodes = {}
         # check that each group has a minimum value
-        for group, nodes in self.nodes.iteritems():
+        for group, nodes in self.nodes.items():
```
riccardomurri: Why? In Python 2, `.items()` takes a *copy* of the list, whereas `.iteritems()` iterates over the original items.

(Not that it matters much, since these lists are small, but I'd like to know the rationale...)

elasticluster/cluster.py
```diff
@@ -601,7 +597,7 @@ def _check_cluster_size(self, min_nodes):
         """
         # check the total sizes before moving the nodes around
         minimum_nodes = 0
-        for group, size in min_nodes.iteritems():
+        for group, size in min_nodes.items():
```
riccardomurri: See above.

elasticluster/cluster.py
```diff
@@ -615,20 +611,20 @@ def _check_cluster_size(self, min_nodes):
 
         # finding all node groups with an unsatisfied amount of nodes
         unsatisfied_groups = []
-        for group, size in min_nodes.iteritems():
+        for group, size in min_nodes.items():
```
riccardomurri: See above.

elasticluster/cluster.py
```diff
@@ -615,20 +611,20 @@ def _check_cluster_size(self, min_nodes):
...
             if len(self.nodes[group]) < size:
                 unsatisfied_groups.append(group)
 
         # trying to move nodes around to fill the groups with missing nodes
         for ugroup in unsatisfied_groups[:]:
             missing = min_nodes[ugroup] - len(self.nodes[ugroup])
-            for group, nodes in self.nodes.iteritems():
+            for group, nodes in self.nodes.items():
```
riccardomurri: See above.

elasticluster/cluster.py
```diff
@@ -811,12 +806,12 @@ def update(self):
...
-                   not (node.preferred_ip and \
-                        node.preferred_ip in node.ips):
-                  node.connect()
+                        not (node.preferred_ip and \
+                                         node.preferred_ip in node.ips):
+                    node.connect()
             except InstanceError as ex:
                 log.warning("Ignoring error updating information on node %s: %s",
-                          node, str(ex))
+                            node, str(ex))
```
riccardomurri: While you're at it, please remove the redundant `str(ex)` -- just `ex` is enough :-)

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+from elasticluster import log
+from elasticluster.exceptions import KeypairError
+
+EXPLICIT_CONFIG = [
+    # 'floating_ip'
+]
+
+
+class LibCloudProvider(AbstractCloudProvider):
+    driver = None
```
riccardomurri: **No,** this creates class-level attributes!

So, of your config has two stanzas with `provider=libcloud` (e.g., one for OpenStack and one for WhateverCloud) then they will share the `driver`, `project_name`, `floating_ip` etc attributes, leading to hard-to-debug issues which depend on the order the configuration stanzas are read into Python memory...

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+    project_name = None
+    floating_ip = False
+    rules = None
+    auth = None
+
+    def pop_args(self, **kwargs):
+        if kwargs.get('username'):
```
riccardomurri: Should be:

    if 'username' in kwargs:
        ...

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+            return [kwargs.pop('username'), kwargs.pop('password')]
```
riccardomurri: If `password` is not provided, then this will return `['foo', None]` which is quite certainly *not* what you want...

If password auth is to be used and a password is not provided, a `ConfigurationError` should be raised.

elasticluster/providers/libcloud_provider.py
riccardomurri: Same in all the other branches of this long `if`.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        elif kwargs.get('access_token'):
+            return kwargs.pop('access token')
+        elif kwargs.get('access_id'):
+            return kwargs.pop('access_id'), kwargs.pop('secret_key')
```
riccardomurri: Same as for `password` above: if `secret_key` is not provided, this will return `[id, None]`

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        elif kwargs.get('service_account_email'):
+            return [kwargs.pop('service_account_email'), kwargs.pop('pem_file')]
```
riccardomurri: Same as for `password` above, if `pem_file` is required then the code should raise `ConfigurationError` if it's missing.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        elif kwargs.get('client_id'):
+            return [kwargs.pop('client_id'), kwargs.pop('client_secret')]
```
riccardomurri: Same as for `password` above, if `client_secret` is required then the code should raise `ConfigurationError` if it's missing.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+class LibCloudProvider(AbstractCloudProvider):
+    driver = None
+    project_name = None
+    floating_ip = False
+    rules = None
+    auth = None
+
+    def pop_args(self, **kwargs):
```
riccardomurri: I would rename it `pop_auth_args` or something similar -- make it clear this is where authentication data is being extracted.  This should also be documented in the method's docstring, which is missing now.

riccardomurri: I would also make this method private since it's basically just auxiliary code for `__init__`, not part of the public API.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+            log.debug("deriving and importing public key from private key")
+            self.__import_pem(key_name, private_key_path, password)
+        elif os.path.exists(os.path.join(self.storage_path, '{}.pem'.format(key_name))):
+            self.__import_pem(key_name, os.path.join(self.storage_path, '{}.pem'.format(key_name)), password)
+        else:
+            with open(os.path.join(self.storage_path, '{}.pem'.format(key_name)), 'w') as new_key_file:
+                new_key_file.write(self.__function_or_ex_function('create_key_pair')(name=key_name))
+            self.__import_pem(key_name, os.path.join(self.storage_path, '{}.pem'.format(key_name)), password)
+
+    """
```
riccardomurri: This should be the docstring for `__import_pem`.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        if not pem:
+            raise KeypairError('could not import %s', pem_file_path)
+        else:
+            f = NamedTemporaryFile('w+t')
+            f.write('{} {}'.format(pem.get_name(), pem.get_base64()))
+            key_import(name=key_name, key_file_path=f.name)
+            f.close()
+
+    """
+    Check if a list function exists for a key on a driver
```
riccardomurri: Misplaced docstring.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    def __check_list_function(self, func):
+        for lf in [getattr(self.driver, c, None) for c in dir(self.driver) if 'list_{}'.format(func) in c]:
```
riccardomurri: I'm not sure what this code is meant to do -- it will return `None` if the `list_...` function is not part of the driver?  What for?

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+            yield lf
+
+    """
+    Check if a function exists for a key on a driver, or if it is an 'extended' function.
```
riccardomurri: Misplaced docstring.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+    :returns a callable or none
+    """
+    def __function_or_ex_function(self, func):
```
riccardomurri: More idiomatic implementation:

```python
def __function_or_ex_function(self, func_name):
    # try function name as given
    try:
        return getattr(self.driver, func_name)
    except AttributeError:
        pass
    # try prefixing name with `ex_`
    try:
        return getattr(self.driver, 'ex_' + func_name)
    except AttributeError:
        pass
    # no such function
    return None`
```

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        if func in dir(self.driver):
+            return getattr(self.driver, func, None), func
+        elif 'ex_{}'.format(func) in dir(self.driver):
+            return getattr(self.driver, 'ex_'.format(func), None)
+        return None
+
+    """
+    Check if a value chain (ex. 'a,b,c') exists in our list of known items (item.name | item.id)
```
riccardomurri: Misplaced docstring.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    @staticmethod
+    def __check_name_or_id(values, known):
+        result = []
```
riccardomurri: I would find the following code easier to read (and it avoids creating N^2
temporary lists)::

```python
def __check_name_or_id(values, known):
    result = []
    for element in values.split(','):
        element = element.strip()
        for item in known:
            if item.id == element or item.name == element:
                result.append(item)
    return result
```

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        except SSHException:
+            try:
+                pem = paramiko.DSSKey.from_private_key_file(os.path.expandvars(os.path.expanduser(pem_file_path)),
+                                                            password)
+            except SSHException:
+                raise KeypairError('could not import %s in rsa or dss format', pem_file_path)
+        if not pem:
+            raise KeypairError('could not import %s', pem_file_path)
+        else:
+            f = NamedTemporaryFile('w+t')
```
riccardomurri: use `with NamedTemporaryFile(...) as f:`

elasticluster/providers/libcloud_provider.py
riccardomurri: Create all instance-level attributes in the `__init__` method.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        options['ex_userdata'] = image_userdata
+        options['username'] = username
+        options['networks'] = options.pop('network_ids', None)
+
+        if self.driver.get_key_pair(key_name):
+            options['auth'] = NodeAuthSSHKey(self.driver.get_key_pair(key_name).public_key)
+        else:
+            options['auth'] = NodeAuthPassword(options.get('image_user_password'))
+
+        for key in [k for k in options.keys() if k not in EXPLICIT_CONFIG]:
```
riccardomurri: `EXPLICIT_CONFIG` is currently the empty list, what's the purpose of this test?

riccardomurri: Anyway, the whole line can be rewritten, possibly in a more readable way, as:

```python
for key in set(options.keys()) - EXPLICIT_CONFIG:
```

(Assuming you make `EXPLICIT_CONFIG` into a set instead of a list.)

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+            list_function = next(self.__check_list_function(key), None)
+            if list_function:
+                populated_list = self.__check_name_or_id(options[key], list_function())
+                if populated_list and len(populated_list) > 0:
```
riccardomurri: It is redundant to write `if populated_list and len(populated_list) > 0`: the two clauses are the same thing.

Just use `if populated_list:`, see e.g. https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+            return []
+        return instance.public_ips + instance.private_ips
+
+    def __prepare_key_pair(self, key_name, private_key_path, public_key_path, password):
+        if not key_name:
+            log.warn('user_key_name has not been defined, assuming password based authentication')
+            return
+
+        if not next(self.__check_list_function('key_pairs'), None):
+            raise KeypairError('key management not supported by provider')
```
riccardomurri: What's the exact meaning of `KeypairError`? These lines seem to use it for "unsupported operation" or possibly as an overload of `ConfigurationError` ("you told me to use a keypair but it's not supported by the backend API").  Then what I am supposed to catch if I want to intercept *data* errors with the keypair? (e.g., it is invalid / file is corrupt / etc.)

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        if not self.__function_or_ex_function('import_key_pair_from_file'):
+            raise KeypairError('key import not supported by provider')
+        if not self.__function_or_ex_function('create_key_pair'):
+            raise KeypairError('key creation not supported by provider')
+
+        if key_name in [k.name for k in next(self.__check_list_function('key_pairs'))()]:
+            log.info('Key pair (%s) is already installed.', key_name)
```
riccardomurri: Since this is just checking the name, I would prefer if the message to the user were like "a keypair by the name `%s` already exists in cloud %s".  If the keypair is there but it refers to a *different* key (e.g., fingerprints don't match) at least we didn't promise anything :-)

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+    def __import_pem(self, key_name, pem_file_path, password):
+        key_import = self.__function_or_ex_function('import_key_pair_from_file')
+        try:
+            pem = paramiko.RSAKey.from_private_key_file(os.path.expandvars(os.path.expanduser(pem_file_path)), password)
+        except SSHException:
+            try:
+                pem = paramiko.DSSKey.from_private_key_file(os.path.expandvars(os.path.expanduser(pem_file_path)),
+                                                            password)
+            except SSHException:
+                raise KeypairError('could not import %s in rsa or dss format', pem_file_path)
```
riccardomurri: This gives no hint as to what has gone wrong.  Isn't it possible to do instead:

```
except SSHException as err:
    raise KeyPairError("could not import {0}, neither as RSA key nor as DSA key: {1}".format(pem_file_path, err))
```

riccardomurri: Anyway, `KeypairError("...", pem_file_path)` is going to bomb. You need a `%` or `.format()` operator there; exception ctors only take a `message` argument.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        if not pem:
+            raise KeypairError('could not import %s', pem_file_path)
```
riccardomurri: See above.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+            self.__import_pem(key_name, os.path.join(self.storage_path, '{}.pem'.format(key_name)), password)
+
+    """
+    Import PEM certificate with provider
+    """
+
+    def __import_pem(self, key_name, pem_file_path, password):
+        key_import = self.__function_or_ex_function('import_key_pair_from_file')
+        try:
+            pem = paramiko.RSAKey.from_private_key_file(os.path.expandvars(os.path.expanduser(pem_file_path)), password)
```
riccardomurri: expression `os.path.expandvars(os.path.expanduser(pem_file_path))` is constant and re-used multiple times; can't you just factor it out into a single variable assignment before the `try:` block?

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+            return getattr(self.driver, func, None), func
+        elif 'ex_{}'.format(func) in dir(self.driver):
+            return getattr(self.driver, 'ex_'.format(func), None)
+        return None
+
+    """
+    Check if a value chain (ex. 'a,b,c') exists in our list of known items (item.name | item.id)
+    """
+    @staticmethod
+    def __check_name_or_id(values, known):
```
riccardomurri: Why `__check_*`? This function *checks* nothing, it gets / filters values...

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+        else:
+            f = NamedTemporaryFile('w+t')
+            f.write('{} {}'.format(pem.get_name(), pem.get_base64()))
+            key_import(name=key_name, key_file_path=f.name)
+            f.close()
+
+    """
+    Check if a list function exists for a key on a driver
+    """
+    def __check_list_function(self, func):
```
riccardomurri: Why `__check_*`? This function *checks* nothing, it iterates over some value.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+            log.debug("importing public key from path %s", public_key_path)
+            key_import = self.__function_or_ex_function('import_key_pair_from_file')
+            if not key_import(name=key_name, key_file_path=os.path.expandvars(os.path.expanduser(public_key_path))):
+                raise KeypairError('failure during import of public key %s', public_key_path)
+        elif private_key_path:
+            if not private_key_path.endswith('.pem'):
+                raise KeypairError('can only work with .pem private keys, derive public key and set user_key_public')
+            log.debug("deriving and importing public key from private key")
+            self.__import_pem(key_name, private_key_path, password)
+        elif os.path.exists(os.path.join(self.storage_path, '{}.pem'.format(key_name))):
```
riccardomurri: `"{}.pem".format(...)` is not valid in Python 2.6 -- you need to rewrite it as `"{0}.pem".format(...)`

See also other occurrences else where in this file.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,210 @@
...
+
+        if key_name in [k.name for k in next(self.__check_list_function('key_pairs'))()]:
+            log.info('Key pair (%s) is already installed.', key_name)
+            return
+
+        if public_key_path:
+            log.debug("importing public key from path %s", public_key_path)
+            key_import = self.__function_or_ex_function('import_key_pair_from_file')
+            if not key_import(name=key_name, key_file_path=os.path.expandvars(os.path.expanduser(public_key_path))):
+                raise KeypairError('failure during import of public key %s', public_key_path)
```
riccardomurri: It would be better if the reason why import has failed were reported...

riccardomurri:
Two additional requests:

- run `pylint` on the `libcloud_provider.py` file and correct errors and warnings it reports.

- I don't understand why `__check_list_function()` should be a generator.  Every time, it's called with `next(..._check_list_function(...), None)` which basically means you retrieve only the first item and discard the rest.  Then just make it a regular function that returns the first match and simplify?

elasticluster/cluster.py
witlox: no python3 support for iteritems

elasticluster/providers/libcloud_provider.py
witlox: actually, this is what I wanted, because I want the library to return the error message, and not write all the plumbing code for the exception handling (which is in libcloud anyway)

elasticluster/providers/libcloud_provider.py
witlox: so within libcloud if you have 'key_pair' functions, there might be a list_key_pairs or ex_list_key_pairs function (or nothing) which allows you to iterate over the key_pairs, this will get the specific function if it exists.

elasticluster/providers/libcloud_provider.py
witlox: so EXPLICIT_CONFIG, is meant for keys which should not be interpreted (with list_* functions) and just be mapped as a plain config kv-pair

elasticluster/exceptions.py
```diff
@@ -25,6 +25,10 @@ class ConfigurationError(Exception):
     pass
 
 
+class UnsupportedError(ConfigurationError):
+    pass
```
riccardomurri: Can you please add a docstring stating what this error is / when it would be raised?  

(I know this is not done for many other exception classes, but it's bad practice that we should aim to improve ;-))

elasticluster/providers/libcloud_provider.py
```diff
@@ -19,47 +19,39 @@
...
 from libcloud.compute.types import NodeState
-from paramiko import SSHException
 
 from elasticluster import AbstractCloudProvider
 from elasticluster import log
-from elasticluster.exceptions import KeypairError
+from elasticluster.exceptions import KeypairError, UnsupportedError
 
-EXPLICIT_CONFIG = [
+EXPLICIT_CONFIG = set(
```
riccardomurri: I still don't get what this is for, since:

- it's a constant
- it's empty...

elasticluster/cluster.py
riccardomurri: fair enough

docs/configure.rst
```diff
@@ -302,6 +302,22 @@ Valid configuration keys for *openstack*
...
+    for all available providers. Configuration keys are provider dependent.
+    For example if you want to configure an azure connection, go to the page
+    https://libcloud.readthedocs.io/en/latest/compute/drivers/azure.html
+    and check what the 'Instantiating a driver' part states, in this case
+    you would need to add the keys 'subscription_id' and 'key_file'.
+
+
+``driver_name``:
+
+    Name of the driver you want to configure (provider you want to connect with).
```
riccardomurri: I would add a link or at least a specification of what exact strings are allowed here.  It should at least mention that this should be the value of libcloud's "provider constant" (which, according to what I can see, is *not* exposed in the documentation so one must look into the source code.

For example: how would I request the "Digital Ocean" driver here?  Should I write "DIGITAL_OCEAN"? or "DigitalOcean"? or what? Does case matter?

This is documentation for end-users: better be explicit.

docs/configure.rst
```diff
@@ -323,6 +339,18 @@ For Google Compute Engine you can use::
...
     gce_project_id=****REPLACE WITH YOUR PROJECT ID****
 
+If you would want to use libcloud to connect to openstack using password authentication
+you can configure the following::
+
+    [cloud/libcloud]
+    provider=libcloud
+    driver_name=openstack
+    auth_url=**** YOUR AUTH URL ****
+    ex_tenant_name=**** YOUR TENNANT NAME ****
```
riccardomurri: `TENNANT` -> `TENANT`

(But note that "tenants" have been renamed to "projects" since a couple of years already, so "newcomers" might not know what a tenant is.)

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,239 @@
...
+        :param options: all the configuration items that should
+                        be forwarded to the driver
+        """
+
+    def __init__(self, driver_name, storage_path=None, **options):
+        self.storage_path = storage_path
+        if 'auth_url' in options and 'ex_force_auth_url' not in options:
+            options['ex_force_auth_url'] = options.get('auth_url').rsplit('/', 1)[0]
+        log.debug('selecting libcloud driver %s', driver_name)
+        driver_class = get_driver(driver_name)
```
riccardomurri: See the above discussion about the "driver_name" config item.  This code seems to require that an internal LibCloud value (= not publicly documented) is used as a configuration item in ElastiCluster.

I would rather provide a way for users to use the "provider constant" (as documented in https://libcloud.readthedocs.io/en/latest/supported_providers.html ) as the driver name.  E.g.::

    from libcloud.compute.types import Provider
    try:
        driver_name = getattr(Provider, driver_name)
    except AttributeError:
        raise ValueError("No libcloud driver for provider {name}".format(name=driver_name))
    driver_class = get_driver(driver_name)

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,239 @@
...
+                f.write('{n} {p}'.format(n=pem.get_name(), p=pem.get_base64()))
+                key_import(name=key_name, key_file_path=f.name)
+
+    def __get_list_function(self, func):
+        """
+        Check if a list function exists for a key on a driver, and if it does, return it
+        :param func: partial function name (ex. key_pair)
+        :return: list function that goes with it (ex. list_key_pairs)
+        """
+        list_functions = [getattr(self.driver, c, None) for c in dir(self.driver) if 'list_{}'.format(func) in c]
```
riccardomurri: This still looks overcomplicated to me.  Have a look at the method names from some providers::

    In [19]: def lf(x):  # print `list_` function names
       ....:     print [attr for attr in dir(x) if 'list_' in attr]
       ....:     

    In [20]: lf(get_driver('digitalocean'))
    ['list_images', 'list_key_pairs', 'list_locations', 'list_nodes', 'list_sizes', 'list_volume_snapshots', 'list_volumes']

    In [21]: lf(get_driver('gce'))
    ['ex_instancegroup_list_instances', 'ex_instancegroupmanager_list_managed_instances', 'ex_list_addresses', 'ex_list_autoscalers', 'ex_list_backendservices', 'ex_list_disktypes', 'ex_list_firewalls', 'ex_list_forwarding_rules', 'ex_list_healthchecks', 'ex_list_instancegroupmanagers', 'ex_list_instancegroups', 'ex_list_instancetemplates', 'ex_list_networks', 'ex_list_project_images', 'ex_list_regions', 'ex_list_routes', 'ex_list_snapshots', 'ex_list_sslcertificates', 'ex_list_subnetworks', 'ex_list_targethttpproxies', 'ex_list_targethttpsproxies', 'ex_list_targetinstances', 'ex_list_targetpools', 'ex_list_urlmaps', 'ex_list_zones', 'list_images', 'list_key_pairs', 'list_locations', 'list_nodes', 'list_sizes', 'list_volume_snapshots', 'list_volumes']

    In [22]: lf(get_driver('rackspace_first_gen'))
    ['ex_list_ip_addresses', 'ex_list_ip_groups', 'list_images', 'list_key_pairs', 'list_locations', 'list_nodes', 'list_sizes', 'list_volume_snapshots', 'list_volumes']

    In [23]: lf(get_driver('ec2'))
    ['ex_list_availability_zones', 'ex_list_internet_gateways', 'ex_list_keypairs', 'ex_list_network_interfaces', 'ex_list_networks', 'ex_list_placement_groups', 'ex_list_reserved_nodes', 'ex_list_route_tables', 'ex_list_security_groups', 'ex_list_subnets', 'list_images', 'list_key_pairs', 'list_locations', 'list_nodes', 'list_regions', 'list_sizes', 'list_snapshots', 'list_volume_snapshots', 'list_volumes']

What I see from this small sample is that:

1. `list_` and `ex_list_` function names are mutually exclusive -- there is never `list_foo` *and* `ex_list_foo`
2. when there is some overlap, LibCloud *changes* the function name slightly (see e.g. `ex_list_keypairs` and `list_key_pairs` in the EC2 provider) so only one of the two would be found by the list comprehension...

I do not really see a reason why `_get_function_or_ex_function` cannot be used instead.

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,239 @@
...
+            return getattr(self.driver, func_name)
+        except AttributeError:
+            pass
+        # try prefixing name with `ex_`
+        try:
+            return getattr(self.driver, 'ex_' + func_name)
+        except AttributeError:
+            pass
+        # no such function
+        return None
```
riccardomurri: Looking at how this function is used, I think the code should raise
`UnsupportedError` here, instead of returning `None`.

The former approach would allow you to get rid of the code at lines 125--130
(you can still pass an optional `errmsg` parameter to customize the exception if
needed for clarity); the latter implies that in principle the return value
should be checked every time, which is cumbersome and error-prone...

elasticluster/providers/libcloud_provider.py
```diff
@@ -0,0 +1,239 @@
...
+    @staticmethod
+    def __get_name_or_id(values, known):
+        """
+        Check if a value chain (ex. 'a,b,c') exists in our list of known items (item.name | item.id), if so, return it.
+        :param values: list of items to check (ex. network_id, etc.)
+        :param known: libcloud items of the named type to validate against
+        :return: list of the libcloud items that match the values (ex. NetworkObject)
+        """
+        result = list()
+        for element in [e.strip() for e in values.split(',')]:
```
riccardomurri: Alternate implementation with linear complexity::

    def __get_name_or_id(values, known):
        candidates = set(value.strip() for value in values.split(','))
        allowed = set(item.id for item in known) + set(item.name for item in known)
        return list(set.intersection(candidates, allowed))

riccardomurri:
I have added some more line-by-line comments.

riccardomurri:
Merged  as e7ff2b1 (squashed many commits into one).

Many thanks for contributing!

PR #436, opened 2017-05-17
Secvol

witlox:
added multiple security groups fix
added ability to boot from volume

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -734,19 +734,19 @@ Optional configuration keys
     Define the type of boot disk to use.
     Only supported when the cloud provider is `google` or `openstack`.
     Supported values are `pd-standard` and `pd-ssd` for Google,
-    or the types available in the openstack volume (cinde) configuration.
+    or the types available in the OpenStack volume (cinder) configuration.
     Default value is `pd-standard` for Google.
-    When using this option for openstack, it creates volumes to be used
+    When using this option for OpenStack, it creates volumes to be used
     as the root disks for the VM's of the specified size, when terminating
     and instance the volume will be deleted automatically. Always specify
-    the boot_disk_size when using this with openstack.
+    the boot_disk_size when using this with OpenStack.
 
 ``boot_disk_size``
     Define the size of boot disk to use.
     Only supported when the cloud provider is `google` or `openstack`.
     Values are specified in gigabytes.
     Default value for Google is 10.
-    No default is given for openstack.
+    No default is given for OpenStack.
 
 ``tags``
     Comma-separated list of instance tags.
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -732,15 +732,21 @@ Optional configuration keys
 
 ``boot_disk_type``
     Define the type of boot disk to use.
-    Only supported when the cloud provider is `google`.
-    Supported values are `pd-standard` and `pd-ssd`.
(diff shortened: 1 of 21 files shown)


riccardomurri:
Please merge the PR I have sent you regarding this one.

docs/configure.rst
```diff
@@ -732,15 +732,21 @@ Optional configuration keys
 
 ``boot_disk_type``
     Define the type of boot disk to use.
-    Only supported when the cloud provider is `google`.
-    Supported values are `pd-standard` and `pd-ssd`.
-    Default value is `pd-standard`.
+    Only supported when the cloud provider is `google` or `openstack`.
+    Supported values are `pd-standard` and `pd-ssd` for Google,
+    or the types available in the openstack volume (cinde) configuration.
```
riccardomurri: cinde -> cinder

docs/configure.rst
```diff
@@ -732,15 +732,21 @@ Optional configuration keys
...
+    the boot_disk_size when using this with openstack.
 
 ``boot_disk_size``
     Define the size of boot disk to use.
-    Only supported when the cloud provider is `google`.
+    Only supported when the cloud provider is `google` or `openstack`.
     Values are specified in gigabytes.
-    Default value is 10.
+    Default value for Google is 10.
+    No default is given for openstack.
```
riccardomurri: "No default ... for openstack" -> so what happens if I do not give a value?

docs/configure.rst
```diff
@@ -732,15 +732,21 @@ Optional configuration keys
...
 ``boot_disk_type``
     Define the type of boot disk to use.
-    Only supported when the cloud provider is `google`.
-    Supported values are `pd-standard` and `pd-ssd`.
-    Default value is `pd-standard`.
+    Only supported when the cloud provider is `google` or `openstack`.
+    Supported values are `pd-standard` and `pd-ssd` for Google,
+    or the types available in the openstack volume (cinde) configuration.
+    Default value is `pd-standard` for Google.
+    When using this option for openstack, it creates volumes to be used
```
riccardomurri: openstack -> OpenStack (here and elsewhere)

witlox:
- no type checking on `boot_disk_size` or `boot_disk_type`: setting defaults invalidates the control flow for the changes, currently there are defaults being set in gce.py, but I cannot think of any descent defaults for OpenStack, for example the name of the volume type is configured per OpenStack instance...

- is there really nothing better than polling, in order to know when a volume is ready? Seeing that the python-cinder client doesn't have any signalling on board, I don't really know of a better solution that won't create a lot of plumbing code.....

riccardomurri:
> > _(RM:)_ no type checking on boot_disk_size or boot_disk_type: 

> _(PW:)_ setting defaults invalidates the control flow for the changes

???

What I mean is: there are lines of code like this:

    bds = int(kwargs['boot_disk_size'])
    
If someone has, e.g., `boot_disk_size=10gb` in the config file, the code above
will trigger a `ValueError` deep down the ElastiCluster call stack in
`OpenStackCloudProvider.start_instance()` which will result in an error message
like `VM "xxx" cannot be started`. A `ConfigurationError` should be the thing
users see instead!

riccardomurri:
Merged into "master" as commit 4373b22 -- thanks a lot for contributing!

Issue #437, opened 2017-05-22
Cannot start nodes on OpenStack cloud using the `libcloud` provider


riccardomurri:
Using the example configuration for using OpenStack with the newly-added
`libcloud` provider, I get the following error::

    2017-05-22 16:31:17 xenia gc3.elasticluster[1925] ERROR Could not start node `master001`: 'str' object has no attribute 'id' -- <type 'exceptions.AttributeError'>
    Traceback (most recent call last):
      File "/home/rmurri/w/elasticluster/elasticluster/cluster.py", line 493, in _start_node
        node.start()
      File "/home/rmurri/w/elasticluster/elasticluster/cluster.py", line 1085, in start
        **self.extra)
      File "/home/rmurri/w/elasticluster/elasticluster/providers/libcloud_provider.py", line 89, in start_instance
        node = self.driver.create_node(**options)
      File "/home/rmurri/.virtualenvs/elasticluster-dev/local/lib/python2.7/site-packages/libcloud/compute/drivers/openstack.py", line 1259, in create_node
        server_params = self._create_args_to_params(None, **kwargs)
      File "/home/rmurri/.virtualenvs/elasticluster-dev/local/lib/python2.7/site-packages/libcloud/compute/drivers/openstack.py", line 1371, in _create_args_to_params
        server_params['imageRef'] = kwargs.get('image').id
    AttributeError: 'str' object has no attribute 'id'

It looks like there's some parameter name mismatch (or a wrongly-adapted
parameter).

witlox:
look like our imageid is not being resolved to an image object, will inspect

witlox:
correction, refactoring made a likely check of list functions to an exact check of list functions, wait with checking need to update code...

witlox:
done

riccardomurri:
Nope... same error.

The issue *may* be that the image ID I have in the config file is invalid and so some critical data structure is filled incorrectly; but this should be caught and properly handled.

PR #438, opened 2017-05-23
fixes for #437

witlox:
missed two formats not supported in python2


riccardomurri:
I have reverted the commit that led to this problem in 56f4cc2 (which basically is the same as this PR).

It does *not* fix #437 though.

PR #442, opened 2017-05-30
Bugfix/ssh host based authentication

HackerMD:


HackerMD:
Somehow the `sshd` service doesn't get restarted. I currently need to manually ssh into all machines and restart `sshd`, otherwise hostbased authentication fails.

Issue #452, opened 2017-06-19
"NotImplementedError: list_key_pairs not implemented for this driver" when using libcloud's `libvirt` driver


riccardomurri:
[Reported][1] by John-Paul Robinson on the ElastiCluster mailing-list:

> It appears the attempt to start a cluster relies on a list_key_pairs() method
> that is not supported in libcloud for the libvirt driver. Unfortunately the docs
> for libcloud's libvirt docs suggest this method is implemented even though the
> support matrix does state the method is not implemented.
> 
> Here's the output from an attempted start of a listed template for the local
> cluster. Note line 146 in `libcloud_provider.py` calls the `list_key_pairs()` method
> which triggers the not implemented section. It's not clear if this is an blocker
> error or if things would still work if the cluster nodes were up an running.
> 
>         (venv) jpr@laptop:~/projects/elasticluster$ elasticluster -c slurmtest/config start slurmtest
>         Starting cluster `slurmtest` with:
>         * 1 frontend nodes.
>         * 2 compute nodes.
>         (This may take a while...)
>         2017-06-07 12:41:21 laptop gc3.elasticluster[20880] ERROR Could not start node `frontend001`: list_key_pairs not implemented for this driver -- <type 'exceptions.NotImplementedError'>
>         Traceback (most recent call last):
>           File "/home/jpr/projects/elasticluster/src/elasticluster/cluster.py", line 493, in _start_node
>             node.start()
>           File "/home/jpr/projects/elasticluster/src/elasticluster/cluster.py", line 1085, in start
>             **self.extra)
>           File "/home/jpr/projects/elasticluster/src/elasticluster/providers/libcloud_provider.py", line 72, in start_instance
>             options.get('image_user_password'))
>           File "/home/jpr/projects/elasticluster/src/elasticluster/providers/libcloud_provider.py", line 146, in __prepare_key_pair
>             if key_name in [k.name for k in list_key_pairs()]:
>           File "/home/jpr/projects/elasticluster/venv/local/lib/python2.7/site-packages/libcloud/compute/base.py", line 1202, in list_key_pairs
>             'list_key_pairs not implemented for this driver')
>         NotImplementedError: list_key_pairs not implemented for this driver
>         ...
> 
> I tried working around this by not giving the key a name which seems like it
> would make `__prepare_key_pair()` skip the initialization steps, but that causes
> parsing issues for the login configuration section in the config.

[...]

PR #477, opened 2017-08-16
452

witlox:
libcloud fixes for #452 (note that @miltondp 's initial configuration was correct)


riccardomurri:
Merged, thanks a lot for contributing this patch!

PR #511, opened 2018-01-02
Bigtop installation fix (and question about spark-thriftserver crash)

pescobar:
I am testing the hadoop+spark setup using ubuntu 16.04. During my first test the bigtop installation was failing so I updated the role to install latest bigtop 1.2.1. This PR includes the modifications I did to install latest bigtop-1.2.1 in ubuntu 16.04

After this change elasticluster goes forward until it crashes in `TASK [spark-common : Install Spark packages (common)]` . It fails trying to start the `spark-thriftserver` service. This is what I get in `/var/log/spark/spark-thriftserver.out` in master node where the service doesn't start:

```
18/01/02 17:01:58 WARN metastore.HiveMetaStore: Retrying creating default database after error: Error creating transactional connection factory
javax.jdo.JDOFatalInternalException: Error creating transactional connection factory
```
```
Caused by: org.datanucleus.exceptions.NucleusException: Attempt to invoke the "BONECP" plugin to create a ConnectionPool gave an error : The specified datastore driver ("org.postgresql.Driver") was not found in the CLASSPATH. Please check your CLASSPATH specification, and the name of the driver.
```

I have checked that deb package `libpostgresql-jdbc-java` is installed in the machine. 

I have no experience with Spark, does anyone knows why this could be failing? Any suggestion about how to fix it? I can try to fix the playbooks but I am not sure how the elasticluster developers would prefer to approach this so I prefer to ask before I do more modifications to try to do a fix which can be merged upstream.

### DIFF
--- a/elasticluster/share/playbooks/roles/bigtop/defaults/main.yml
+++ b/elasticluster/share/playbooks/roles/bigtop/defaults/main.yml
@@ -2,7 +2,7 @@
 ---
 
 # What release of Apache Bigtop to get packages from
-bigtop_release: '1.1.0'
+bigtop_release: '1.2.1'
 
 # identifier for GnuPG key(s) used to sign packages from the Apache Bigtop repo
 bigtop_signing_key:
--- a/elasticluster/share/playbooks/roles/bigtop/defaults/main.yml
+++ b/elasticluster/share/playbooks/roles/bigtop/defaults/main.yml
@@ -3,8 +3,3 @@
 
 # What release of Apache Bigtop to get packages from
 bigtop_release: '1.2.1'
-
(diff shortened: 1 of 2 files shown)


riccardomurri:
I can confirm that the playbooks stop when installing `spark-thriftserver` because of an error in the Debian package post-install script.

riccardomurri:
The problem seems to be that `spark-thriftserver` is started too early: when the error reported occurs, ElastiCluster's configuration for Spark has not yet been deployed so (in particular) the PostGreSQL Java connector cannot be found (it's not in the default Java CLASSPATH, for whatever reason).

I have a fix -- I'm merging this into "master" and then committing the fix there.

PR #504, opened 2017-11-10
Add support for GPUs on Google Cloud

riccardomurri:

PR #512, opened 2018-01-03
add support for gpus on GCE

dkoes:
This allows the user to include requests for GPUs on nodes in the flavor field for GCE.  It will need to be changed once GPUs are out of beta.  I don't know the syntax to provide similar functionality for other cloud providers.


riccardomurri:
Isn't this already covered by https://github.com/gc3-uzh-ch/elasticluster/pull/504 ? (Which I'm going to merge as soon as I can get some feedback from testers.)

dkoes:
Sure looks like it.

PR #528, opened 2018-03-13
Make HTCondor playbook work with Debian 8.x and Ubuntu 14.04.

riccardomurri:
(Debian 9 ("stretch") and Ubuntu 16.04 ("xenial") are not supported
upstream.)

In addition:

- Make `htcondor` into a proper Ansible role.

- Rename `condor_workers` to `condor_worker` to use the same naming
  convention that we use for other batch-queuing systems.

Many thanks to @maikenp for suggesting the needed fixes!


riccardomurri:
@maikenp This works for me -- I will merge as soon as you can confirm that it works!

Issue #529, opened 2018-03-14
Azure support not working


riccardomurri:
It looks like Azure support is currently broken both in the `azure` provider and in the `libcloud` one.
For more details and discussion, see: https://groups.google.com/forum/#!topic/elasticluster/yp_YMJPF3u8

dsteinkraus:
Commented in the linked forum. (I didn't know there was a libcloud provider, so I have nothing to contribute on that.)

Smahane:
@riccardomurri  Thanks for announcing this issue here. It was driving me crazy! any promised date of when this bug could get fixed? I'm hoping to be able to create an Azure cluster later this week.

riccardomurri:
> ny promised date of when this bug could get fixed?

@Smahane The basic functionality is now available in PR #530 and works.  I'm polishing it and will be ready to merge by end of this week into "master" (hence, in the Docker image).  You're welcome to test the PR already and tell me if there are any macroscopic bugs or missing functionality ;-)

Smahane:
Great! I will give it a try. What is the Docker image? Any docs?

riccardomurri:
New Azure provider has now been merged into "master"; in a few minutes the Docker image will be rebuilt with it as well.

@Smahane The Docker image is meant to provide a 0-install experience for ElastiCluster (through Docker). Here's the relevant docs: http://elasticluster.readthedocs.io/en/latest/install.html#quickstart

PR #530, opened 2018-03-17
The new Azure backend.

riccardomurri:
This is meant to solve the issues raised in #529 

**Note:** the configuration for Azure has changed!

An example configuration file with the new keys/format is available at: https://github.com/riccardomurri/elasticluster/blob/42ad511df66aa8d14bad267c649cf41ba8aa53b6/examples/slurm-and-gridengine-on-azure-complete.conf


riccardomurri:
What is currently working (at least in my setup):

- starting a cluster
- stopping a cluster (removes, node, its NIC, and the public IP address)

What is still left:

- removing the "storage account" used for the node
- honoring the "security_group" configuration setting (assuming it makes sense on Azure)
- removing the VNet, Subnet, and Resource Group when last node is deleted.

monajemi:
@riccardomurri I have tested this PR. Everything works fine so far that I have created CPU instances.
Stop does not work properly as you mentioned. I will try building a GPU instance too soon and will let you know. 

Thanks,
Hatef

riccardomurri:
Hi @monajemi many thanks for testing!  Can you please pull/update your code before you test GPUs?  With the latest commit 0024266 deletion works good for me.

monajemi:
@riccardomurri 
certificate must be removed from azure schema. It creates an error if not there. 

```
ERROR` In section `cloud/azure`: Missing keys: 'certificate'
```

riccardomurri:
@monajemi Thanks for pointing out the certificate issue; newest code now rejects it and exists with error if `certificate=...` is in the config file. (Too much has changed in the Azure config to stay backwards compatible.)

riccardomurri:
Just merged into "master" -- many thanks to @monajemi for testing!

monajemi:
@riccardomurri It appears that for some instances (e.g., NC instances), you can only use them 
if the storage type is HDD. If you choose SDD, you will get an error. I know for sure that NC instances have to be on HDD. I was on a call with Azure now. ND instances can be setup with both.

riccardomurri:
@monajemi I'm a bit confused: you start the comment with NC instances, then conclude with ND instances... Is it a misspelling?

monajemi:
Hi @riccardomurri No, It is not. Azure offers NC and ND instances for GPUs. https://docs.microsoft.com/en-us/azure/virtual-machines/windows/sizes-gpu

The NC instances can only be launched if HDD is used for VM Disk Type. 
The ND instances are older and can be launched with both SSD and HDD. 

Because I could not initiate NC instance using elasticluster, I had to switch to ND, and I am waiting for them to increase my ND limit.

I can build NC instances via the portal though. I think in EC, one has to link NC instance to HDD Disk, otherwise the python API gives error.

Hope this makes more sense now. Azure is confusing! :)

Issue #545, opened 2018-04-04
Local variable 'public_key_content' referenced before assignment


ltupin:
Hello,

When I try to deploy a SLURM cluster on GCP (with the Cloud Shell VM ), I get the following error:

```
ERROR Could not start node MyServerName001: local variable 'public_key_content' referenced before assignment -- <type 'exceptions.UnboundLocalError'>
Traceback (most recent call last):
  File "/home/google324748_student/elasticluster2/src/elasticluster/cluster.py", line 542, in _start_node
    node.start()
  File "/home/google324748_student/elasticluster2/src/elasticluster/cluster.py", line 1186, in start
    **self.extra)
  File "/home/google324748_student/elasticluster2/src/elasticluster/providers/gce.py", line 321, in start_instance
    "value": "%s:%s" % (username, public_key_content),
UnboundLocalError: local variable 'public_key_content' referenced before assignment'>
```
Environment:

```
uname -a
Linux cs-6000-devshell-vm-e190fe0f-5d5e-4830-9e98-8fb2ee2de1b3 3.16.0-5-amd64 #1 SMP Debian 3.16.51-3+deb8u1 (2018-01-08) x86_64 GNU/Linux

date
Wed Apr  4 15:03:38 CEST 2018

git branch
* master

elasticluster --version
elasticluster version 1.3.dev2
```

I found this error were related to ./elasticluster/providers/gce.py file.

To resolve it I simply add the variable declaration used to construct the body in  the previous block as is:  

```
        elif tags is not None:
            raise TypeError(
                "The `tags` argument to `gce.start_instance`"
                " should be a string or a list, got {T} instead"
                .format(T=type(tags)))
        
        with open(public_key_path, 'r') as f:
            public_key_content = f.read()

        compute_metadata = [
            {
                "key": "ssh-keys",
                "value": "%s:%s" % (username, public_key_content),
            },
```

PR #546, opened 2018-04-04
Local variable referenced before assignment #545

ltupin:
I report details in the issue #545. It resolve the problem for my use-case.


ltupin:
You're welcome !

PR #554, opened 2018-04-26
Recreate known_hosts file when calling cluster.remove_node()

kysolvik:
As described in #176, currently removing nodes does not remove their entries in known_hosts. This can cause connection failures if a new node reuses the old IP address.

To fix this, cluster.remove_node() calls cluster._gather_node_ip_addresses() with a new "remake" flag set to True, which removes the self.known_hosts_file so that it will be recreated from scratch with only the remaining nodes.

I'm not sure if this is the best approach, but it does recreate the known_hosts file without the removed node. Could also directly remove the known_hosts_file in cluster.remove_node() rather than having the "remake" flag.

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -385,6 +385,10 @@ class Cluster(Struct):
                     node.stop()
                 self._naming_policy.free(node.kind, node.name)
                 self.repository.save_or_update(self)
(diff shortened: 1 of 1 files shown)


riccardomurri:
Excellent! Many many thanks for contributing! It's great to see old bug being fixed :-)

kysolvik:
Happy to help!

PR #555, opened 2018-05-03
add task to start kubeadm services on Debian nodes

gmauro:
I added a task to ensure that services needed by kudeadm (Docker and Kubelet)  start on Debian nodes


riccardomurri:
Looks good to me, merging.

Thanks so much for contributing to ElastiCluster!

PR #557, opened 2018-05-09
Kubernetes: deploy the dashboard, set a version and have configuration easily accessible  from normal user

gmauro:
Several improvements:
- a role to deploy the Dashboard into the Kubernetes cluster just created
- Kubernetes version to install: default is the latest but can be set a different one
- Kubernetes cluster configuration is copied into a /opt/k8s/config to allow normal user to easily access it


elasticluster/share/playbooks/roles/kubernetes-common/tasks/debian.yml
```diff
@@ -22,6 +22,18 @@
     - kubeadm
     - kubectl
     - kubernetes-cni
+  when: not kubernetes_version.apt.install
+
+- name: Install packages - fixed version
+  apt:
+    name: "{{ item }}"
+    allow_unauthenticated: '{{ not insecure_https_downloads|default("no")|bool }}'
```
riccardomurri: Re: `allow_unauthenticated: ...` -- isn't there an extra `not` in the "value" part?

IOW, shouldn't it be::

    allow_unauthenticated: '{{ insecure_https_downloads|default("no")|bool }}'

elasticluster/share/playbooks/roles/kubernetes-common/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+kubernetes_version:
+  apt:
+    label: 1.10.2-00
+    install: false
```
riccardomurri: Missing end-of-line at EOF

elasticluster/share/playbooks/roles/kubernetes-apps/tasks/dashboard.yml
```diff
@@ -0,0 +1,22 @@
...
+    - create_ClusterRoleBinding.yaml
+  when: kubernetes_dashboard_create_admin_account
+
+- name: Create ServiceAccount
+  command: kubectl --kubeconfig /etc/kubernetes/admin.conf apply -f /var/tmp/create_ServiceAccount.yaml
+  when: kubernetes_dashboard_create_admin_account
+
+- name: Create ClusterRoleBinding
+  command: kubectl --kubeconfig /etc/kubernetes/admin.conf apply -f /var/tmp/create_ClusterRoleBinding.yaml
+  when: kubernetes_dashboard_create_admin_account
```
riccardomurri: Missing end-of-line and EOF

elasticluster/share/playbooks/roles/kubernetes-apps/tasks/dashboard.yml
```diff
@@ -0,0 +1,22 @@
+# See https://github.com/kubernetes/dashboard
+#     https://github.com/kubernetes/dashboard/wiki/Creating-sample-user
+
+- name: Deploy Kubernetes Dashboard
+  command: kubectl --kubeconfig /etc/kubernetes/admin.conf  apply -f {{ kubernetes_dashboard_configuration }}
```
riccardomurri: I know that unconditionally calling `kubectl` is already been done in other parts of the playbook, but what happens if the playbook is re-run after a cluster has been set up?  Will it fail? (app/user already exists -- cannot create) or exit successfully? (app/user already exists, nothing to do)

elasticluster/share/playbooks/roles/kubernetes.yml
```diff
@@ -8,6 +8,19 @@
     - master
   roles:
     - kubernetes-master
+  post_tasks:
```
riccardomurri: `post_tasks:` could be just `tasks:`, for consistency with other playbooks.

elasticluster/share/playbooks/roles/kubernetes.yml
```diff
@@ -8,6 +8,19 @@
...
+      path: "/opt/k8s"
+      state: directory
+      mode: 0755
+
+  - name: Copy configuration into kube directory to allow normal user to access it
+    copy:
+      remote_src: yes
+      src: /etc/kubernetes/admin.conf
+      dest: "/opt/k8s/config"
+      mode: 0644
```
riccardomurri: I would make this `0444` -- there's little point in anyone modifying that file so better raise a warning.

elasticluster/share/playbooks/roles/kubernetes.yml
```diff
@@ -8,6 +8,19 @@
...
   roles:
     - kubernetes-master
+  post_tasks:
+  - name: Create kube directory for normal user
+    file:
+      path: "/opt/k8s"
+      state: directory
+      mode: 0755
+
+  - name: Copy configuration into kube directory to allow normal user to access it
```
riccardomurri: I'm curious: what's the reason for allowing any user to read a *copy* of the K8s config?  If there are no secrets in it, why not let users read `/etc/kubernetes/admin.conf` directly?  If there are secrets, why allow users to read them?  (Authorized admins can still do it via `sudo`)

riccardomurri:
This is my major concern with this PR:

> I know that unconditionally calling kubectl is already been done in other parts of the playbook, but what happens if the playbook is re-run after a cluster has been set up? Will it fail? (app/user already exists -- cannot create) or exit successfully? (app/user already exists, nothing to do)

(Not a blocker, I am not sure I understand the motivation behind it.)

riccardomurri:
Also this:

> I'm curious: what's the reason for allowing any user to read a copy of the K8s config? If there are no secrets in it, why not let users read /etc/kubernetes/admin.conf directly? If there are secrets, why allow users to read them? (Authorized admins can still do it via sudo)

(Not a blocker, I am not sure I understand the motivation behind it.)

elasticluster/share/playbooks/roles/kubernetes.yml
gmauro: https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#instructions
In step 2/4, you can see:
"...
Your Kubernetes master has initialized successfully!

To start using your cluster, you need to run (as a regular user):

  mkdir -p $HOME/.kube
  sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
  sudo chown $(id -u):$(id -g) $HOME/.kube/config
..."

/etc/kubernetes/admin.conf can be read only by by root or by sudo user.
In situation as mine, where I am controlling the cluster from my notebook, you need to have a local copy of that file and do an scp as regular user is quite better.
The point here is that I have mimic

elasticluster/share/playbooks/roles/kubernetes-common/defaults/main.yml
gmauro: I can take care of this

elasticluster/share/playbooks/roles/kubernetes-apps/tasks/dashboard.yml
gmauro: I can take care of this

elasticluster/share/playbooks/roles/kubernetes-apps/tasks/dashboard.yml
gmauro: At the moment re-run of the kubernetes playbooks fail at the "Join Kubernetes cluster" step of the workers, before of Dashboard's deployment.
I agree with you, would be better to add a check to skip this step (and others similar) but I haven't a solution ready right now. Any hints?

elasticluster/share/playbooks/roles/kubernetes.yml
gmauro: the idea was to emphasize that this task have to happen after the role, but could be tasks as well.
I can take care of this.

elasticluster/share/playbooks/roles/kubernetes-common/tasks/debian.yml
gmauro: yes. I will remove it.

elasticluster/share/playbooks/roles/kubernetes-apps/tasks/dashboard.yml
riccardomurri: > Any hints?

Is there a `kubectl` invocation that would list the entities you're going to create?  If yes, you could do something like:

    -name: create stuff
     shell: |
          if ! (kubectl list-stuff | grep -q "stuff name"); then
               kubectl ... apply -f ...
           fi

But don't waste too much time on this -- I know the K8s playbook is
already *not* idempotent so if you can fix this quickly OK, otherwise
it goes onto the pile of "future enhancements".

riccardomurri:
Merged, thanks a lot! especially for bearing with my amendment requests :-)

Looking forward to your next improvement!

PR #561, opened 2018-05-16
add tag to filter tasks related to glusterfs mountpoint

gmauro:
In this way, we can setup a node as glusterfs-client without the need to Mount GlusterFS filesystem at the same time.


riccardomurri:
Merged, many thanks for this addition!

PR #562, opened 2018-05-17
verify if the node has been added to a cluster

gmauro:
adding a little bit of idempotence to the K8s playbooks.


riccardomurri:
LGTM, thanks again!

PR #564, opened 2018-06-03
Hpc-extra - new playbook

dpshelio:
I'm not sure whether this is the way to add new playbooks. But since I've done it and it adds a couple of dependencies I needed I believe it's better to share it with whoever may need it.

This adds a new module on openMPI3 for centos7 to `hpc-common` and a new playbook: `hpc-extra`. Maybe this type of setup would be better to do via easy-build?

I'm happy for this to be closed if this is not the right way to do this.

### DIFF
--- a/elasticluster/share/playbooks/roles/hpc-common/templates/etc/modulefiles/Core/mpi/openmpi3.lua.j2
+++ b/elasticluster/share/playbooks/roles/hpc-common/templates/etc/modulefiles/Core/mpi/openmpi3.lua.j2
@@ -1,0 +1,58 @@
+--
+-- THIS FILE IS CONTROLLED BY ELASTICLUSTER
+-- local modifications will be overwritten
+-- the next time `elasticluster setup` is run!
+--
+
+--
+-- OpenMPI module for use with 'environment-modules' package
+--
+help([[
+Load the version of OpenMPI3 that came with the base OS.
+
+Note that it is configured (via env. variables) to use TCP sockets for
+message passing, and avoid even probing for IB.
+
+To find out more about OpenMPI, visit: http://www.open-mpi.org/
+]])
+whatis("Load the version of OpenMPI3 that came with the base OS.")
+
+-- no two MPI packages can be loaded at the same time
+family("MPI")
+
+{% if ansible_os_family == 'Debian' -%}
+-- (e.g. the Python `openmpi` package) were removed.
+-- NOTE: I've not found openmpi3 for debian like as a separate package;
+-- This hasn't been tested!
+--
+{%- elif ansible_os_family == 'RedHat' -%}
+prepend_path("PATH", 	        "/usr/lib64/openmpi3/bin")
+prepend_path("LD_LIBRARY_PATH", "/usr/lib64/openmpi3/lib")
+prepend_path("INCLUDE_PATH", "/usr/include/openmpi3-x86_64")
+prepend_path("PYTHONPATH",      "/usr/lib64/python2.7/site-packages/openmpi3")
+prepend_path("MANPATH",         "/usr/share/man/openmpi3-x86_64")
+setenv("MPI_BIN",		"/usr/lib64/openmpi3/bin")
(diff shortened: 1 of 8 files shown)


riccardomurri:
Hello @dpshelio, thank a lot for taking the time to contribute to
ElastiCluster!

In general the PR looks good to me, but I have a general question: Why
is "hpc-extra" a separate playbook from "hpc-common"?  It looks to me
that it provides more recent versions of popular HPC packages on
CentOS7 -- why not just modify `hpc-common` to provide those?

I'll add minor issues as line-by-line comments.

elasticluster/share/playbooks/roles/hpc-common/templates/etc/modulefiles/Core/mpi/openmpi3.lua.j2
```diff
@@ -0,0 +1,58 @@
...
+To find out more about OpenMPI, visit: http://www.open-mpi.org/
+]])
+whatis("Load the version of OpenMPI3 that came with the base OS.")
+
+-- no two MPI packages can be loaded at the same time
+family("MPI")
+
+{% if ansible_os_family == 'Debian' -%}
+-- (e.g. the Python `openmpi` package) were removed.
+-- NOTE: I've not found openmpi3 for debian like as a separate package;
```
riccardomurri: Line 25: If no OpenMPI3 packages are available for Debian/Ubuntu, this section should be removed and just fall into the `else` branch.

(BTW, Debian has an `openmpi3` package, but it's only in the "buster"/testing distribution; you may want to check https://pkgs.org/download/openmpi to see cross-distribution support for a package.)

elasticluster/share/playbooks/roles/hpc-extra/tasks/init-Debian.yml
```diff
@@ -0,0 +1,14 @@
+---
+
+- name: Load distribution-specific defaults (Debian/Ubuntu)
+  set_fact:
+    hpc_common_packages:
```
riccardomurri: Why (re)set this variable here?  For influencing what the `hpc-common` role does, it's too late (it has already run as a dependency), for use in the `hpc-extra` role, it should rather be named `hpc_extra_packages`.

elasticluster/share/playbooks/roles/hpc-extra/tasks/init-Debian.yml
```diff
@@ -0,0 +1,14 @@
+---
+
+- name: Load distribution-specific defaults (Debian/Ubuntu)
```
riccardomurri: This really looks a copy+paste of the corresponding file in `hpc-common`; if there is nothing to do additionally for Debian/Ubuntu, just do nothing :-)

elasticluster/share/playbooks/roles/hpc-extra/tasks/init-RedHat.yml
```diff
@@ -0,0 +1,24 @@
...
+  set_fact:
+    hpc_extra_packages:
+      - centos-release-scl
+      - devtoolset-7
+      - devtoolset-7-gcc-c++
+      - devtoolset-7-gcc-gfortran
+      - cmake3
+      - automake
+      - autoconf
+      - tmux
```
riccardomurri: Line 13--15: `tmux`, `git`, `wget` should all have already been installed by the `common` role.

dpshelio:
@riccardomurri - I put it separate because I was learning my way through ansible and I wanted to be sure I was breaking the right stuff :).

If you think these are of use as in common I'm happy to move them to common.

I'll update the PR with your suggestions soon.

riccardomurri:
> I'll update the PR with your suggestions soon.

Thanks!  Note that I'll be on vacation for the coming two weeks, so probably not very responsive :-)

riccardomurri:
I think the gist of this went into commit c5133953 -- please re-open or make a new PR if you think there is something that still needs to be updated.

PR #569, opened 2018-06-25
Pre-emptible instances cannot be automatically restarted

gillesdegottex:
From the doc: https://cloud.google.com/compute/docs/instances/preemptible
"Preemptible instances cannot live migrate or be set to automatically restart when there is a maintenance event."
Since the state of "when there is a maintenance event" cannot be known at instance creation, I guess the best is to set automaticRestart to False when preemptible.
(GCE always rejected by instance creation when automaticRestart=True and preemptible=True anyway)

### DIFF
--- a/elasticluster/providers/gce.py
+++ b/elasticluster/providers/gce.py
@@ -440,6 +440,9 @@ class GoogleCloudProvider(AbstractCloudProvider):
             # see: https://cloud.google.com/compute/docs/gpus#restrictions
             instance['scheduling']['onHostMaintenance'] = 'TERMINATE'
             instance['scheduling']['automaticRestart'] = True
+            if scheduling_option['preemptible']:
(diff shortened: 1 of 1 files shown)


riccardomurri:
Hello @gillesdegottex, many thanks for contributing this fix!

A question: the code in this PR will only run if GPU accelerators are requested, but the combination of scheduling=preemptible and automaticRestart seems more general than that.  Is there a reason why the `if` is only applied to clusters with GPUs?

gillesdegottex:
no one, my case was for GPU accelerated instance, so I didn't think further :)
fixed in 1778584

riccardomurri:
Squashed and merged (with little edits) as commit b33b1c35

Many thanks for fixing this and reacting so quickly to my comments! :-)

PR #570, opened 2018-06-26
Propagate "SSH proxy command" settings to Ansible.

### LINKED ISSUE #567: SSH ProxyCommand not passed to Ansible
When `ssh_proxy_command` is given in a `[cluster/...]` section in order to facilitate access via a bastion host, it is not propagated into the Ansible configuration, resulting in a failure to access machines.

The workaround is to specify `ansible_ssh_args=-o ProxyCommand='...'` in the `[setup/...]` section, but this should really be done automatically, especially as the `setup` sections are meant to be generic, and the `ProxyCommand` encodes a particular bastion host, which might change per-cluster.

riccardomurri:
Fixes #567

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -52,6 +52,7 @@ from elasticluster.exceptions import (
 from elasticluster.repository import MemRepository
 from elasticluster.utils import (
     Struct,
+    expand_ssh_proxy_command,
     get_num_processors,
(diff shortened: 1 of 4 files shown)


riccardomurri:
Needs testing.

riccardomurri:
Works for me, merging into "master" since no-one available to test.

PR #577, opened 2018-08-07
Openstack: support for anti-affinity server groups.

flashvoid:
Hello,

We made this feature for one of our clients and interested if it would be welcomed upstream.

The goal is to provide better results for cpu hungry applications on openstack clouds with cpu overcommit.

### DIFF
--- a/elasticluster/providers/openstack.py
+++ b/elasticluster/providers/openstack.py
@@ -129,6 +129,8 @@ class OpenStackCloudProvider(AbstractCloudProvider):
         or `None` (default, meaning try v3 first and fall-back to v2).
     :param cacert: Path to CA certificate bundle (for verifying HTTPS sessions)
         or ``None`` to use the systems' default.
+    :param bool manage_anti_affinity_groups: use openstack aaf to spread the load
+    :param str anti_affinity_group_prefix: prefix for creating server groups
     """
 
     __node_start_lock = threading.Lock()
@@ -149,6 +151,8 @@ class OpenStackCloudProvider(AbstractCloudProvider):
                  # this is deprecated in favor of `compute_api_version`
                  nova_api_version=None,
                  cacert=None,  # keep in sync w/ default in novaclient.Client()
+                 manage_anti_affinity_groups=False,
+                 anti_affinity_group_prefix=None
     ):
         # OpenStack connection params
         self._os_auth_url = self._get_os_config_value('auth URL', auth_url, ['OS_AUTH_URL']).rstrip('/')
@@ -188,6 +192,10 @@ class OpenStackCloudProvider(AbstractCloudProvider):
         self.request_floating_ip = request_floating_ip
         self._instances = {}
         self._cached_instances = {}
+
+        self.anti_affinity_group_prefix = anti_affinity_group_prefix
+        # constant for anti-affinity routines.
+        self.middle_token = ".number."
 
     @staticmethod
     def _get_os_config_value(thing, value, varnames, default=_NO_DEFAULT):
@@ -499,10 +507,38 @@ class OpenStackCloudProvider(AbstractCloudProvider):
                         .format(id=volume.id, delete_on_terminate=1)),
             }
 
-        # due to some `nova_client.servers.create()` implementation weirdness,
-        # the first three args need to be spelt out explicitly and cannot be
-        # conflated into `**vm_start_args`
-        vm = self.nova_client.servers.create(node_name, image_id, flavor, **vm_start_args)
+
+
+        out_of_capacity = 'No valid host was found. There are not enough hosts available.'
+
+        # create anti-affinity groups for spawning servers to ensure server spread
+        # equally across hosts in the cloud.
+        # This will create a server group, spawn hosts in the group until it's full
+        # and then create a new group.
+        if self.anti_affinity_group_prefix:
+            with OpenStackCloudProvider.__node_start_lock:
+
+                group = self._get_current_anti_affinity_group()
+                vm_start_args['scheduler_hints']={ 'group' : group.id }
+                vm = self.nova_client.servers.create(node_name, image_id, flavor, **vm_start_args)
+                self._wait_for_status(vm, ["ACTIVE", "ERROR"], 10)
+
+                if vm.status == 'ERROR' and vm.fault['message'] == out_of_capacity:
+                    log.debug("Deleting instance %s(%s) in group %s", vm.name, vm.id, group.name)
+                    self.nova_client.servers.delete(vm.id)
+
+                    group=self._get_next_anti_affinity_group()
+                    vm_start_args['scheduler_hints']={ 'group' : group.id }
+
+                    vm = self.nova_client.servers.create(node_name, image_id, flavor, **vm_start_args)
+
+                log.debug("Started server %s in group %s", vm.name, group.name)
+        else:
+            # due to some `nova_client.servers.create()` implementation weirdness,
+            # the first three args need to be spelt out explicitly and cannot be
+            # conflated into `**vm_start_args`
+            vm = self.nova_client.servers.create(node_name, image_id, flavor, **vm_start_args)
+
 
         # allocate and attach a floating IP, if requested
         if self.request_floating_ip:
@@ -529,6 +565,7 @@ class OpenStackCloudProvider(AbstractCloudProvider):
         self._init_os_api()
         instance = self._load_instance(instance_id)
         instance.delete()
+        self._delete_anti_affinity_groups()
         del self._instances[instance_id]
 
     def get_ips(self, instance_id):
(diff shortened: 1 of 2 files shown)


elasticluster/providers/openstack.py
```diff
@@ -499,10 +507,38 @@ def start_instance(self, key_name, public_key_path, private_key_path,
...
+        # and then create a new group.
+        if self.anti_affinity_group_prefix:
+            with OpenStackCloudProvider.__node_start_lock:
+
+                group = self._get_current_anti_affinity_group()
+                vm_start_args['scheduler_hints']={ 'group' : group.id }
+                vm = self.nova_client.servers.create(node_name, image_id, flavor, **vm_start_args)
+                self._wait_for_status(vm, ["ACTIVE", "ERROR"], 10)
+
+                if vm.status == 'ERROR' and vm.fault['message'] == out_of_capacity:
```
riccardomurri: Is it possible that the `out_of_capacity` message gets localized and we miss it?  Is there an error code, exception class, or other more invariant means of checking what the issue is?

elasticluster/providers/openstack.py
```diff
@@ -499,10 +507,38 @@ def start_instance(self, key_name, public_key_path, private_key_path,
...
+                    vm_start_args['scheduler_hints']={ 'group' : group.id }
+
+                    vm = self.nova_client.servers.create(node_name, image_id, flavor, **vm_start_args)
+
+                log.debug("Started server %s in group %s", vm.name, group.name)
+        else:
+            # due to some `nova_client.servers.create()` implementation weirdness,
+            # the first three args need to be spelt out explicitly and cannot be
+            # conflated into `**vm_start_args`
+            vm = self.nova_client.servers.create(node_name, image_id, flavor, **vm_start_args)
```
riccardomurri: This branch of the `if` misses the `with ...__node_start_lock`

elasticluster/providers/openstack.py
```diff
@@ -555,6 +592,39 @@ def is_instance_running(self, instance_id):
 
     # Protected methods
 
+    def _delete_anti_affinity_groups(self):
+        for group in self._list_anti_affinity_groups():
+            self.nova_client.server_groups.delete(group.id)
+
+    def _get_next_anti_affinity_group(self):
+        group_name=self._make_next_group_name(self._get_current_anti_affinity_group().name)
+        return self.nova_client.server_groups.create(name=group_name, policies='anti-affinity')
```
riccardomurri: Will this work for *any* OpenStacl user?  Or does it require admin access?

elasticluster/providers/openstack.py
```diff
@@ -555,6 +592,39 @@ def is_instance_running(self, instance_id):
...
+
+    def _get_current_anti_affinity_group(self):
+        groups=self._list_anti_affinity_groups()
+
+        if groups:
+            group = sorted(groups, key=lambda g: g.name)[-1]
+        else:
+            group = self.nova_client.server_groups.create(name=self.anti_affinity_group_prefix + self.middle_token + "1", policies='anti-affinity')
```
riccardomurri: Is the `"1"` really a constant here?  If yes, why not make it part of `self.middle_token`?

elasticluster/providers/openstack.py
```diff
@@ -189,6 +193,10 @@ def __init__(self, username, password, project_name, auth_url,
         self._instances = {}
         self._cached_instances = {}
 
+        self.anti_affinity_group_prefix = anti_affinity_group_prefix
+        # constant for anti-affinity routines.
+        self.middle_token = ".number."
```
riccardomurri: Very minor point: I would prefer if `self.middle_token` was `self._middle_token` so it does not look like it's part of the public API / parameters of the class.  (I know the ElastiCluster sources are inconsistent on this :-))

riccardomurri:
Thank you very much for this contibution!  It would be fantastic to have it merged upstream; I think it would be extremely useful to whoever is running on mixed-workload oversubscribed clouds.

The code looks good to me; I have one important question: will it work for *any* OpenStack user or do anti-affinity groups require some kind of admin privileges?

I have noted a few other minor points in the sources; great if you have time to fix (or talk back) yourself, otherwise I can take care of that when merging.

Thanks again for taking the time to contribute!

elasticluster/providers/openstack.py
flashvoid: This is a very good point, unfortunately, there doesn't seem to be any way to assert the reason of an error other than the error message. The call is async, there is no distinct error code and error itself is similar to one when an entire cloud runs out of resources.

Maybe it makes sense to drop second conditional and just leave as it is. The logic would be to stop growing the group after first instance that fails to start, regardless of a reason.

elasticluster/providers/openstack.py
flashvoid: No that was on purpose.
`Cluster.start()` allows to modes of node creation, sync and async. Async isn't compatible with aaf groups because we need feedback from the previous instance before making a decision about next.

So effectively, `anti_affinity_group_prefix` implies `-p 1`, and this line represents backward compatibility for people who want async and do not want to use aaf groups.

P.S. I would, however, make a point that async in case of openstack doesn't save much time, `create_instance` API is async by nature so `-p 1` wouldn't make much difference.

elasticluster/providers/openstack.py
flashvoid: No admin privileges needed in our cloud and as far as I can tell it is a default behavior to allow regular users to use this feature.

elasticluster/providers/openstack.py
flashvoid: It is a constant and it isn't a part of `self.middle_token` because former is used in `self._list_anti_affinity_groups()` for substring matching.

elasticluster/providers/openstack.py
flashvoid: ok, will do

elasticluster/providers/openstack.py
flashvoid: Just made an extra test, thats what openstack tells when scheduler fails to schedule an instance due to anti-affinity.
```
ipdb> pp vm.fault
{u'code': 500,
 u'created': u'2018-08-08T04:22:56Z',
 u'message': u'No valid host was found. There are not enough hosts available.'}
```

riccardomurri:
Thank you very much for following up!  I'll be back on Friday: then I'll test the code and merge.

riccardomurri:
So the code looks fine and doesn't break existing functionality :-) I cannot really test the AAF feature, as it is not supported by our local OpenStack; I'll take your word that it works :-)

A couple more questions related to AAF:

1. at the moment, the AAF prefix is set cloud-wide; does it make sense to set it cluster-wide?  (i.e., different clusters on the same cloud use different prefixes)  Otherwise, it seems to me that stopping a cluster will could tear down the AAF groups used for *other* clusters.

2. I still don't get the prefix + middle token + number thing: I understand that you want to list all AAF groups "belonging" to an ElastiCluster session (hence the need for a *unique* substring, which at the moment is played by prefix+middle), and that you need to generate multiple AAF groups (hence the addition of the final number).  But then why is the middle token constant? if it's constant, it plays no uniqueness role and can be omitted; if it serves the purpose of distinguishing ElastiCluster-generated AAF groups from ones that users may legitimately create on their own, then we should probably chose something more specific than `.number.` (e.g., `_elasticluster_` to indicate origin, or some randomly-generated string to minimize probability of collisions).

flashvoid:
@riccardomurri 

> I cannot really test the AAF feature, as it is not supported by our local OpenStack; I'll take your word that it works :-)
I will be happy to set up an account for you on our public cloud so you could test things. Once we work out the reviews here. 

> then we should probably choose something more specific than .number.
I agree, will fix the middle token to be `_elasticloud_`

> at the moment, the AAF prefix is set cloud-wide; does it make sense to set it cluster-wide?
Well, I was addressing specific use-case and might have missed the bigger picture (after all that what reviews are for :-) ). What would be the desired way to pass cluster-wide configuration down to the openstack provider? would you recommend passing it with the Node object?

riccardomurri:
Hello again!

Sorry this had completely fallen off my radar -- I have restructured the code a bit to work with parallel starting of instances, and for using a single configuration item (boolean `use_anti_affinity_groups`):

https://github.com/gc3-uzh-ch/elasticluster/compare/master...riccardomurri:pr/%23577?expand=1

I think we still need to record the AAF group names somehow; I understand "delete all" will only be performed by OpenStack if all server groups are empty, but I think it would be cleaner if every instance just cleaned up its own server group.

riccardomurri:
>  I think it would be cleaner if every instance just cleaned up its own server group.

This is now implemented in my `pr/#577` branch, on top of your commits.  Would you have time to test it?  I do not have access to an OpenStack installation with AAF groups.

Issue #588, opened 2018-11-19
if 1 cluster .yml file is not valid (e.g. empty), elasticluster stops reading the others as well.


smaffiol:
how to reproduce:
in elasticluster storage have an empty <cluster>.yml file
Note: this can be produced if, for exaple, the setup process was interrupted before elasticluster was able to fill the content.
then run:
$ elasticluster list
gc3.elasticluster[7408] ERROR Error: type object argument after ** must be a mapping, not NoneType

What should be the expected behavior ?
elasticluster should report the failure of the import of the specific cluster but continue in the process with the others.

S.

riccardomurri:
Fixed in #589, thanks @smaffiol !

PR #589, opened 2018-11-20
Sergio: patch  for issue588:

smaffiol:
Handle the case when a cluster .yml file is empty.
Elasticluster reports the error but continues with the other clusters.


riccardomurri:
Fixes #588

riccardomurri:
Thanks, merged!

PR #603, opened 2019-02-12
sergio: fix for issue 602

smaffiol:
set optionxform to `str` to avoind config keys being lowercased.

S.


riccardomurri:
Already merged as part of #603

PR #605, opened 2019-02-14
Enable clusters to be paused and restarted.

### LINKED ISSUE #583: Pause GCE cluster instead of destroying it on "stop"
A nice feature would be the ability to properly pause a GCE (+maybe other backends) cluster. "elasticluster stop x" currently destroys the instances + disks, which means that significant setup time is required even for small-sized clusters (100+ compute nodes). Since disk space is relatively cheap in comparison to compute time, it should be sufficient in many cases to stop the instances instead of destroying them. This can be done manually already, but seems to lead to some minor(?) issues, such as that the original IPs, which change upon stop/start are incorrect. Related to issue https://github.com/gc3-uzh-ch/elasticluster/issues/505.

ndmckinley:
This has five core changes.
- adds `pause` command, which stores a dictionary of information about nodes.
- adds `resume` command, which takes that dictionary and starts the node back up.
- implements both of those for GCE.
- creates a top-level playbook (`resume.yml`) which runs after restart.
- fixes a couple bugs in ansible playbooks I found while testing it.

This fixes #583.

There's a couple things it doesn't do yet - it doesn't have control of thread count plumbed through to the command line, for instance - but I wanted to get it out there for a first pass of review.

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -748,7 +748,7 @@ class Cluster(Struct):
         """Pause all VMs in this cluster and store data so that they
         can be restarted later.
         """
-        log.debug("Pausing cluster `%s`", self.name)
+        log.info("Pausing cluster `%s` ...", self.name)
         failed = self._pause_all_nodes()
         if os.path.exists(self.known_hosts_file):
             os.remove(self.known_hosts_file)
@@ -765,7 +765,7 @@ class Cluster(Struct):
         """
         Resume all paused VMs in this cluster.
         """
-        log.debug("Resuming cluster `%s`", self.name)
+        log.info("Resuming cluster `%s` ...", self.name)
         failed = self._resume_all_nodes()
         for node in self.get_all_nodes():
             node.update_ips()
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -779,6 +779,12 @@ class Cluster(Struct):
                     "restarted.  Check error messages above and consider "
                     "re-running `elasticluster resume %s` if "
                     "necessary.", self.name)
+            return
+        self._run_resume_provisioner()
+        if not self._setup_provider.resume_cluster(self):
+            log.warning("Elasticluster was not able to guarantee that the "
+                        "cluster restarted correctly - check the errors "
+                        "above and check your config.")
 
     def _delete_saved_data(self):
         self._setup_provider.cleanup(self)
@@ -861,6 +867,9 @@ class Cluster(Struct):
         for node_name in successfully_resumed:
             del self.paused_nodes[node_name]
         return len(self.paused_nodes)
+
+    def _run_resume_provisioner(self):
+        pass
 
     def get_ssh_to_node(self, ssh_to=None):
         """
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -823,14 +823,7 @@ class Cluster(Struct):
                     node.name, node.instance_id, err, err.__class__)
         return failed
 
-    def _pause_all_nodes(self, max_thread_pool_size=0):
-        """Pause all cluster nodes - ensure that we store data so that in
-        the future the nodes can be restarted.
-
-        :return: int - number of failures.
-        """
-        failed = 0
-
+    def _make_thread_pool(self, max_thread_pool_size):
         try:
             if max_thread_pool_size == 0:
                 max_thread_pool_size = 4 * get_num_processors()
@@ -839,6 +832,16 @@ class Cluster(Struct):
                 "Cannot determine number of processors!"
                 " will start nodes sequentially...")
             max_thread_pool_size = 1
+        thread_pool_size = min(len(self.get_all_nodes()), max_thread_pool_size)
+        return Pool(processes=thread_pool_size)
+
+    def _pause_all_nodes(self, max_thread_pool_size=0):
+        """Pause all cluster nodes - ensure that we store data so that in
+        the future the nodes can be restarted.
+
+        :return: int - number of failures.
+        """
+        failed = 0
 
         def _pause_specific_node(node):
             if not node.instance_id:
@@ -857,8 +860,7 @@ class Cluster(Struct):
                 return None
 
         nodes = self.get_all_nodes()
-        thread_pool_size = min(len(nodes), max_thread_pool_size)
-        thread_pool = Pool(processes=thread_pool_size)
+        thread_pool = self._make_thread_pool(max_thread_pool_size)
         for node, state in zip(nodes, thread_pool.map(_pause_specific_node, nodes)):
             if state is None:
                 failed += 1
@@ -871,16 +873,7 @@ class Cluster(Struct):
         if not self.paused_nodes:
             log.warning("Didn't find any paused nodes - not resuming anything.")
             return
-        try:
-            if max_thread_pool_size == 0:
-                max_thread_pool_size = 4 * get_num_processors()
-        except RuntimeError:
-            log.warning(
-                "Cannot determine number of processors!"
-                " will start nodes sequentially...")
-            max_thread_pool_size = 1
-        thread_pool_size = min(len(self.paused_nodes), max_thread_pool_size)
-        thread_pool = Pool(processes=thread_pool_size)
+        thread_pool = self._make_thread_pool(max_thread_pool_size)
 
         def _resume_single_node(node_name):
             node_state = self.paused_nodes[node_name]
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -780,7 +780,6 @@ class Cluster(Struct):
                     "re-running `elasticluster resume %s` if "
                     "necessary.", self.name)
             return
-        self._run_resume_provisioner()
         if not self._setup_provider.resume_cluster(self):
             log.warning("Elasticluster was not able to guarantee that the "
                         "cluster restarted correctly - check the errors "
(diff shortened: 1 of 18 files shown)


elasticluster/cluster.py
```diff
@@ -740,6 +744,47 @@ def stop(self, force=False, wait=False):
...
+                    "stopped.  Some nodes may still be running - "
+                    "check error messages above and consider "
+                    "re-running `elasticluster pause %s` if "
+                    "necessary.", self.name)
+
+    def resume(self):
+        """
+        Resume all paused VMs in this cluster.
+        """
+        log.debug("Resuming cluster `%s`", self.name)
```
riccardomurri: I'd raise this to severity `info`:
```suggestion
        log.info("Resuming cluster `%s` ...", self.name)
```

elasticluster/cluster.py
```diff
@@ -740,6 +744,47 @@ def stop(self, force=False, wait=False):
         else:
             self._delete_saved_data()
 
+    def pause(self):
+        """Pause all VMs in this cluster and store data so that they
+        can be restarted later.
+        """
+        log.debug("Pausing cluster `%s`", self.name)
```
riccardomurri: I'd raise this to severity `info`:
```suggestion
        log.info("Pausing cluster `%s` ...", self.name)
```

elasticluster/cluster.py
```diff
@@ -740,6 +744,47 @@ def stop(self, force=False, wait=False):
...
+        failed = self._resume_all_nodes()
+        for node in self.get_all_nodes():
+            node.update_ips()
+        self._gather_node_ip_addresses(
+            self.get_all_nodes(), self.start_timeout, self.ssh_probe_timeout)
+        self.repository.save_or_update(self)
+        if failed:
+            log.warning(
+                    "Not all cluster nodes have been successfully "
+                    "stopped.  Some nodes may still be running - "
```
riccardomurri: It looks like the message has been copied from `pause()` but not edited in full? e.g., `stopped` should be `resumed`? `running` could be `off`?

elasticluster/share/playbooks/resume.yml
```diff
@@ -0,0 +1,58 @@
+---
+- name: Prepare VM for running Ansible
```
riccardomurri: I think we can skip this task in the "resume" playbook -- we know that the "setup" playbook has run at least once, so Python definitely *is* there.

elasticluster/share/playbooks/resume.yml
```diff
@@ -0,0 +1,58 @@
...
+
+- import_playbook: roles/slurm-resume.yml
+#
+# These cluster types have been confirmed not to need
+# additional customization.
+# - import_playbook: roles/gridengine-resume.yml
+# - import_playbook: roles/hadoop-resume.yml
+# 
+# Remaining cluster types may or may not work after resuming.
+#
```
riccardomurri: If we import/run `before.yml`, then we should also import/run `after.yml` here.

elasticluster/share/playbooks/roles/jupyter/tasks/python.yml
```diff
@@ -39,7 +39,7 @@
         # PostGreSQL
         - libpq-dev
         # MySQL compatible
-        - libmysqlclient-dev
+        - default-libmysqlclient-dev
```
riccardomurri: I need to check if this is available on all supported Debian and Ubuntu versions...

elasticluster/share/playbooks/roles/slurm-resume.yml
```diff
@@ -0,0 +1,46 @@
+---
+
+- name: Restart SLURMcltd
+  hosts: slurm_master
+  tasks:
+    - service:
+        name=slurmctld
```
riccardomurri: Please prefer the YAML syntax to the old Ansible syntax (here and in the following lines).
```suggestion
        name: slurmctld
```

(I know it's still used in some playbooks... takes time to update all legacy code...)

elasticluster/share/playbooks/roles/jupyter/tasks/python.yml
ndmckinley: The issue is that `libmysqlclient-dev` is not available on debian stretch.  This one is.  Which versions are supported?  I'm happy to go look at those.

elasticluster/cluster.py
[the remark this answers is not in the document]
ndmckinley: Oh, I see - I missing deleting line 778! Thanks!

elasticluster/cluster.py
```diff
@@ -779,6 +824,79 @@ def _stop_all_nodes(self, wait=False):
...
+        try:
+            if max_thread_pool_size == 0:
+                max_thread_pool_size = 4 * get_num_processors()
+        except RuntimeError:
+            log.warning(
+                "Cannot determine number of processors!"
+                " will start nodes sequentially...")
+            max_thread_pool_size = 1
+
+        def _pause_specific_node(node):
```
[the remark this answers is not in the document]
ndmckinley: I chose to put it here for readability reasons, preferring a little bit of locality over a little bit of performance.  Happy to change it - it's a very mild preference of mine.

elasticluster/cluster.py
```diff
@@ -1205,6 +1323,16 @@ def stop(self, wait=False):
...
             self.instance_id = None
 
+    def pause(self):
+        """
+        Pause the VM  instance and return the info needed to restart it.
+        """
+        if self.instance_id is None:
+            raise ValueError("Trying to stop unstarted node.")
+        resp = self._cloud_provider.pause_instance(self.instance_id)
+        self.update_ips()
```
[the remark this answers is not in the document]
ndmckinley: That is its effect on GCP, but I'm not familiar with all cloud providers and suspected there might be some nuance there.  We can save a good 30 seconds if I can just set `self.preferred_ip` to `None` right here.

elasticluster/share/playbooks/resume.yml
ndmckinley: Done.

elasticluster/share/playbooks/resume.yml
ndmckinley: Done.

ndmckinley:
Did a rebase to grab changes in master.

ndmckinley:
One more minor issue with ubuntu - systemd doesn't kill the processes appropriately on Ubuntu 18 LTS.  Fixed it in this last push.  I think that's everything.

elasticluster/cluster.py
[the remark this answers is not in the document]
riccardomurri: I think it's safe to just set `self.preferred_ip` to `None` here: we're pausing the node.

It is important that `self.update_ips()` is called on resume, though.

elasticluster/share/playbooks/roles/slurm-resume.yml
```diff
@@ -0,0 +1,56 @@
...
+        path: /var/run/slurm-llnl
+        state: directory
+      when: 'is_debian_compatible and (is_debian_8_or_later or is_ubuntu_15_10_or_later)'
+
+    - name: Get running slurm processes
+      shell: "ps -ef | grep -v grep | grep -w slurmd | awk '{print $2}'"
+      register: running_slurm
+      when: 'is_debian_compatible and (is_debian_8_or_later or is_ubuntu_15_10_or_later)'
+    - name: Kill running slurm processes
+      shell: "kill -9 {{ item }}"
```
riccardomurri: I think you can condense lines 32--39 into:

```
- name: Kill all running SLURM processes
  command: 'killall -w -g {{item}}'
  loop:
    - /usr/bin/slurmd
    - /usr/bin/slurmctld
```

ndmckinley: This is only the worker, so I only need to kill `slurmd`, not `slurmctld`.  `slurmctld` shuts down cleanly on the master with `service`, so this isn't necessary there.  Done, testing it now.

elasticluster/cluster.py
```diff
@@ -779,6 +824,79 @@ def _stop_all_nodes(self, wait=False):
...
+                max_thread_pool_size = 4 * get_num_processors()
+        except RuntimeError:
+            log.warning(
+                "Cannot determine number of processors!"
+                " will start nodes sequentially...")
+            max_thread_pool_size = 1
+        thread_pool_size = min(len(self.paused_nodes), max_thread_pool_size)
+        thread_pool = Pool(processes=thread_pool_size)
+
+        def _resume_single_node(node_name):
```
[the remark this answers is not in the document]
ndmckinley: Done.

elasticluster/share/playbooks/roles/slurm-resume.yml
ndmckinley: Seems to work.

elasticluster/share/playbooks/roles/jupyter/tasks/python.yml
ndmckinley: It's available in ubuntu bionic, cosmic, and disco.  The old name, `libmysqlclient-dev`, is in all versions of ubuntu.  `default-` is in debian jessie, stretch, buster, and sid, and non-`default-` is not in any debian.  Looks like we need to use `default-` in debian and non-`default-` in ubuntu.

riccardomurri: Thanks a lot for taking the time to check on all distros!  I'm making a PR with the change now.

riccardomurri:
I think we're good to merge now?

elasticluster/share/playbooks/roles/jupyter/tasks/python.yml
ndmckinley: I've got that in another PR I'm getting ready to send.  :)

ndmckinley:
Think so!  :)

riccardomurri:
Merged into "master" now.

Many many thanks for this effort and for the quick response to all my remarks! Hope this is one of many great PRs :-)

PR #610, opened 2019-03-01
Change the name of the mysql client dev library chosen for Debian / Ubuntu.

ndmckinley:
This should make sure that the correct mysql client is installed in either Debian or Ubuntu.

### DIFF
--- a/elasticluster/share/playbooks/roles/jupyter/tasks/python.yml
+++ b/elasticluster/share/playbooks/roles/jupyter/tasks/python.yml
@@ -38,8 +38,6 @@
       Debian:
         # PostGreSQL
         - libpq-dev
-        # MySQL compatible
-        - libmysqlclient-dev
       RedHat:
         - postgresql-devel
     jupyter_additional_python_packages:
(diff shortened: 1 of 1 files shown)


ndmckinley:
I'm testing this one now - It takes a while to create the cluster to use this, so it may be a few minutes before I'm 100% sure this works.

elasticluster/share/playbooks/roles/jupyter/tasks/python.yml
```diff
@@ -54,6 +52,17 @@
     state: '{{ pkg_install_state }}'
   loop: '{{jupyter_additional_os_packages[ansible_os_family]}}'
 
+- name: Install mysql development package.
+  package:
+    name: 'mysqlclient-dev'
```
riccardomurri: Shouldn't this be `libmysqlclient-dev`?
```suggestion
    name: 'libmysqlclient-dev'
```

elasticluster/share/playbooks/roles/jupyter/tasks/python.yml
```diff
@@ -54,6 +52,17 @@
...
+    state: '{{pkg_install_state}}'
+  when: is_ubuntu_15_10_or_later
```
riccardomurri: I think this would work on Ubuntu 14.04 as well.
```suggestion
  when: is_ubuntu_14_04_or_later
```

riccardomurri:
How about this one instead? https://github.com/riccardomurri/elasticluster/commit/6469804b74b72c8a59adad578b05db0ed68f4eba

ndmckinley:
Yep, yours is better.  I'll get ansible-idiomatic configs figured out evenutally!  Thanks!

PR #612, opened 2019-03-06
Fix templating error

loveshack:
Ansible complained at the original.  (I thought I'd already reported this...)


riccardomurri:
I don't think that line could lead to a templating error --  at worst a warning to the extent that Jinja expressions should not be used with `when:`...

Anyway: merged, it's indeed better with your fix. Many thanks for taking the time to contribute!

Issue #611, opened 2019-03-05
deployment is painfully slow


loveshack:
Deploying on AWS is painfully slow for me.  It takes ~30 minutes for two nodes on c5.large instances with current development source and an AWS HPC AMI, i.e. with various relevant packages already installed.

Sitting watching, it seemed most of that is in package updates and installation, which just shouldn't take that long (and doesn't, comparing with a simple vagrant setup, for instance).  I discovered that's because of loops over packages. I modified at least the most important instances of loops to use a package list instead.  That reduced the time to ~10 minutes.  I could supply changes, but I'm not an ansible expert, so probably haven't done it appropriately.

riccardomurri:
If you are installing from source, can you please install Python package [`ara`][1] (`pip install ara`) into the same virtualenv where ElastiCluster is running?  That will create a SQLite DB with data about every Ansible task executed, which I can then use for profiling.

Don't hold great expectations though :-( Ansible is quite slow especially in copying files and installing packages so we won't be able to shave much.  On the plus side, though, setup time is the same for any number of nodes up to `ANSIBLE_FORK`, so you get ~30 minutes for 2 through 2*num_cpus nodes...

[1]: https://pypi.org/project/ara/

loveshack:
> If you are installing from source, can you please install Python
> package [`ara`][1] (`pip install ara`) into the same virtualenv where
> ElastiCluster is running?  That will create a SQLite DB with data
> about every Ansible task executed, which I can then use for profiling.

I can do, but is there much point without fixing the factor of three
from package update/installation that's obvious by eye?  Also I don't
know how much variation there might be in the timing, though it seemed
reasonably consistent in two or three goes.

> Don't hold great expectations though :-( Ansible is quite slow
> especially in copying files and installing packages so we won't be
> able to shave much.

I noticed it's slow, but the main slowness in package installation seems
to be how it's done.  (Even keeping loops, there's redundancy from
mentioning some redundantly in different places and others that are in
the base distribution; also some seem odd choices for a default
installation.)

riccardomurri:
Commits b704938..d6f312e8 reap the low-hanging fruit of using lists instead of loops and turning off checksums in existence checks.

Does this give enough speed-up for you?

loveshack:
For what it's worth, I tried, but I can't get ara to work according to the instructions.  It always gives an error like this, independent of ARA_LOG_FILE.

```
[...]

PR #614, opened 2019-03-06
Use list of packages, not (slow!) loop

loveshack:
This is what I did about the slowness of package handling  (#611), though it may not be the appropriate way to do it.


riccardomurri:
Many thanks for starting this! Indeed "loop squashing" was removed from Ansible in 2018 and now passing a package list is the recommended (and speedier!) way to go...

Building on your PR, I have changed all the other apt/yum/package invocations and will merge this PR as soon as I am done testing the changes.

riccardomurri:
Merged (with additions) in commits b704938..c1c43b7 (note that your code shows up as authored by `root`, not sure if you're happy with it)

Many thanks for contributing!

PR #615, opened 2019-03-06
Set user agent to elasticluster when making requests to gce.

ndmckinley:
Note: I believe that this doesn't introduce a new dependency - apiclient depends on googleapiclient, so we should always have this module imported.  In any event, it runs fine without any additional packages in a virtualenv that only contains elasticluster, so I feel pretty confident in it.

### DIFF
--- a/elasticluster/providers/gce.py
+++ b/elasticluster/providers/gce.py
@@ -43,6 +43,7 @@ from oauth2client.file import Storage
 from oauth2client.client import OAuth2WebServerFlow
 from oauth2client.tools import run_flow
 from oauth2client.tools import argparser
+import pkg_resources
 
 # Elasticluster imports
 from elasticluster import log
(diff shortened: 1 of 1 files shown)


riccardomurri:
> I believe that this doesn't introduce a new dependency - apiclient depends on googleapiclient, so we should always have this module imported.

"Explicit is better than implicit" (cit.) ;-)

Can you pls add the dependency in `setup.py` so we're safe if `apiclient` changes its dependency model?

Also: do you want to include ElastiCluster's version in the User Agent string?

ndmckinley:
Good news, I checked and the necessary dependency is already in there.  :)

Good call on the version - added that.

riccardomurri:
Squashed and merged in commit 6e2b6eb (had to do manually since GitHub complained about merge conflicts).

Thanks for contributing!

Issue #455, opened 2017-06-26
No authentication secret should be in the config file(s)


riccardomurri:
Posting configuration files is sometimes necessary for (remote) debugging, but
sanitizing them to remove passwords and other authentication information can be
easily forgotten (one is focused on solving a problem, not on possible security
implications).

So: authentication secrets should be written into a separate file, which should
*never* be posted.

- OpenStack can read authentication information from the environment (see,
  e.g., <https://wiki.openstack.org/wiki/OpenStackClient/Authentication>);
  the [`os-client-config`][1] library supports splitting configuration data into
  a (non-sensitive) `clouds.yaml` file and a (protected) `secure.yaml` file
  (see: <https://docs.openstack.org/developer/os-client-config/#splitting-secrets>

- Google's cloud libraries also allow use of a separate file for auth info,
  shared by all GCP client apps;
  see: <https://github.com/gc3-uzh-ch/elasticluster/issues/369>
  
- I am unsure how things work / should work with AWS. (Is it just fine to have very restricted IAM roles?)

- Apache LibCloud provides more low-level facilities, so in this case it's
  completely up to ElastiCluster to design a way of protecting auth info or
  splitting it into a separate "secrets" file.

[1]: https://docs.openstack.org/developer/os-client-config/

verdurin:
Consider using Ansible Vault for secrets?

riccardomurri:
> Consider using Ansible Vault for secrets?

Would not be an option here: policy is that local secrets should never ever leave the campus' network, no matter if encrypted.

In addition, this would anyway require that people somehow import their secrets into a vault -- I would like to use the cloud providers' "native" method if possible.

Issue #369, opened 2016-12-23
Use `GOOGLE_APPLICATION_CREDENTIALS` env var for locating GCE access credentials


riccardomurri:
Google's SDK documentation recommends storing access credentials in a JSON file, whose location is pointed to by the `GOOGLE_APPLICATION_CREDENTIALS` env var:
https://developers.google.com/identity/protocols/application-default-credentials#howtheywork

ElastiCluster should:

1. allow the use of a JSON credentials file pointed to from ElastiCluster's config file, instead of the triplet `gce_client_id`/`gce_client_secret`/`gce_project_id` as we do now.

2. locate the credentials file using the said env. var. in case the location is not specified in the config file.

riccardomurri:
It's possible that 2. comes for free by just using Google's GCE Python libraries.

riccardomurri:
Actually, it would be nice to use any authentication system that `gcloud` uses so people can use the `elasticluster ` CLI interchangeably with `gcloud compute`.

riccardomurri:
Fixed in PR #616 by @ndmckinley -- thanks a lot Nathan!

Issue #327, opened 2016-10-17
Feature Request: Google cloud authentication


NathanRVance:
When running Elasticluster from a google cloud instance, it's difficult without a GUI to accept Google's authentication prompt via a web browser. Cloud instances already have access to Google Cloud APIs, so would it be possible to configure Elasticluster to use that rather than a client id and secret?

NathanRVance:
Just to be clear, when I provide the following entries in the configuration file:
> [cloud/google]
> provider=google
> gce_project_id=****REPLACE WITH YOUR PROJECT ID****
> gce_client_id=****REPLACE WITH YOUR CLIENT ID****
> gce_client_secret=****REPLACE WITH YOUR SECRET KEY****

and I start elasticluster on a google cloud computer (not my local computer) with:
> $ elasticluster start ipython -n mycluster

I get command line output stating that elasticluster is opening a particular URL in my webbrowser. Since my headnode is in a cloud computer, I don't have a webbrowser, authentication can't be completed, and the process hangs.

When setting up a cloud computer I've selected the option to give it unlimited access to GCE APIs. I can verify that I can spawn new instances from the command line using either google's REST API or the gcloud command line utility. Is it possible to bypass this web based authentication in elasticluster and use the authentication already in place?

riccardomurri:
> When setting up a cloud computer I've selected the option to give it unlimited
> access to GCE APIs. I can verify that I can spawn new instances from the
> command line using either google's REST API or the gcloud command line
> utility. Is it possible to bypass this web based authentication in
> elasticluster and use the authentication already in place?

Sounds like a reasonable feature request, and something ElastiCluster
should indeed implement.

But I do not know Google's GCE API well enough to tell whether this is 
actually possible or not. Perhaps @mbookman can comment on the feasibility?

mbookman:
[gcloud](https://cloud.google.com/sdk/gcloud/) now provides the ability for you to create credentials to be used by other applications.

The command to do this is:

    gcloud auth application-default login

The docs for the command are [here](https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login).

This allows you to explicitly create [Application Default Credentials](https://developers.google.com/identity/protocols/application-default-credentials). That page has an [example of picking up ADC from Python](https://developers.google.com/identity/protocols/application-default-credentials#callingpython).

[...]

Issue #250, opened 2016-05-15
Programmatic authentication


adraghici:
Hi all,

I am wondering whether there is any way to perform the authentication step programatically (i.e. without needing to follow the URL and entering the verification code manually)?

My problem is that I am trying to find a way to start and manage an Elasticluster GCE fleet from machines where installing Elasticluster itself is not an option.

I am doing this by bootstrapping a GCE instance with Elasticluster installed, which acts as a coordinator node that can take commands to interact with the cluster. However, since I want to automate the creation of the cluster, I need to be able to bypass the manual authentication step and do it in some other way.

I did some reading and looked into the source code and it seems like this is all the basic workflow for OAuth 2.0 for applications accessing Google resources, so it might not be really possible to hack it.

TL;DR: I'm trying to fully automate the deployment of a GCE cluster on behalf of a user who only provides his credentials, but I am currently blocked by the requirement for user interaction imposed by OAuth.

Any ideas or suggestions are much appreciated!

Thanks!

riccardomurri:
Hello Adrian,

I am not an expert with GCE, but they seem to have a notion of "service accounts" for this purpose: https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances

Have you tried asking the same question on the GCE forums?

Ciao,
Riccardo

adraghici:
I read about service accounts, but I was wondering if it was possible to do it somehow wit OAuth since this is the way Elasticluster does authentication. I might try to add support for doing this using service accounts over the following days, so I'll keep this issue open for now.

Thanks!

riccardomurri:
I think this is now fixed in PR #616 by @ndmckinley  -- please re-open if you think that doesn't cover your use-case.

PR #616, opened 2019-03-06
Add the ability to use gcloud credentials or other Application Default Credentials

ndmckinley:
With this change, folks don't have to do anything complicated on headless machines - any time they're able to use `gcloud` or any other ADC-using tool, they should be able to use `elasticluster`.

This should effectively resolve the issues in https://github.com/gc3-uzh-ch/elasticluster/issues/369, https://github.com/gc3-uzh-ch/elasticluster/issues/327, https://github.com/gc3-uzh-ch/elasticluster/issues/250, as well as the GCE portion of https://github.com/gc3-uzh-ch/elasticluster/issues/455.

### DIFF
--- a/elasticluster/conf.py
+++ b/elasticluster/conf.py
@@ -198,9 +198,9 @@ CLOUD_PROVIDER_SCHEMAS = {
 
     'google': {
         "provider": 'google',
-        "gce_client_id": nonempty_str,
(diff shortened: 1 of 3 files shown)


riccardomurri:
Very very useful work, thanks a lot!

PR #617, opened 2019-03-08
Make travis show green for external contributors.

ndmckinley:
If the build isn't allowed access to secure variables, don't try to run the docker build.


riccardomurri:
Very, very useful -- thanks a lot!

PR #623, opened 2019-03-29
Be compatible with Python3

riccardomurri:


riccardomurri:
I think this is ready to merge, as soon as someone else can provide some confirmation that it works not just on my machine.

riccardomurri:
It looks like there is some issue in Azure support:

```
>>> from azure.mgmt.compute.models import DiskCreateOption
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rmurri/.virtualenvs/elasticluster-py3/lib/python3.6/site-packages/azure/mgmt/compute/models.py", line 10, in <module>
    from .v2018_09_01.models import *
ModuleNotFoundError: No module named 'azure.mgmt.compute.v2018_09_01'
```

I'm not sure whether this is a problem with Azure's Python packages, or the way they are required/imported in ElastiCluster.

elasticluster/cluster.py
```diff
@@ -184,7 +189,7 @@ def __init__(self, name, user_key_name='elasticluster-key',
 
         # attributes that have already been defined trump whatever is
         # in the `extra` dictionary
-        for key in extra.keys():
+        for key in list(extra.keys()):
```
yarikoptic: I don't think that `list()` is needed here (and there is a similar case below)
```suggestion
        for key in extra.keys():
```

Having said that, unrelated to py2to3 conversion, I would recommend to do first `extra = extra.copy()` right before the loop or instead of modifying `extra` in-place, create a new dict to populate on inverse condition (`if not hasattr(self.key)`) to populate `self.extra`.  The reason is that if someone uses a dictionary outside to initiate extra (e.g. to reuse the same parameters), you will be causing a side-effect here, e.g.
```
extra = dict(name=...)
c1 = Cluster(..., **extra)
# there will be no "name" in extra any more ATM
```

elasticluster/cluster.py
```diff
@@ -1264,7 +1276,7 @@ def __init__(self, name, cluster_name, kind, cloud_provider, user_key_public,
         self.preferred_ip = extra.pop('preferred_ip', None)
         self.ips = extra.pop('ips', [])
         # Remove extra arguments, if defined
-        for key in extra.keys():
+        for key in list(extra.keys()):
```
yarikoptic: the same comment here on not needing `list()` and causing a side-effect

elasticluster/conf.py
```diff
@@ -790,7 +807,7 @@ def _cross_validate_final_config(objtree, evict_on_error=True):
         valid = True
         # ensure all cluster node kinds are defined in the `setup/*` section
         setup_sect = cluster['setup']
-        for groupname, properties in cluster['nodes'].items():
+        for groupname, properties in list(cluster['nodes'].items()):
```
yarikoptic: another not needed `list()`?

elasticluster/conf.py
```diff
@@ -816,7 +833,7 @@ def _cross_validate_final_config(objtree, evict_on_error=True):
         # EC2-specific checks
         if cluster['cloud']['provider'] == 'ec2_boto':
             cluster_uses_vpc = ('vpc' in cluster['cloud'])
-            for groupname, properties in cluster['nodes'].items():
+            for groupname, properties in list(cluster['nodes'].items()):
```
yarikoptic: `list()` to go?

elasticluster/migration_tools.py
```diff
@@ -76,15 +81,15 @@ def __setstate_upgrade__(self, state):
     if 'thread_pool_max_size' not in state:
         self._patches['thread_pool_max_size'] = (NotPresent(), 10)
 
-    for attr, values in self._patches.items():
+    for attr, values in list(self._patches.items()):
```
yarikoptic: `list()` to go away?

elasticluster/migration_tools.py
```diff
@@ -112,20 +117,20 @@ def execute(self):
...
+
         patch_cluster()
         for cluster in clusters:
             print("Cluster `%s`" % cluster)
             print("path: %s" % repo.storage_path + '/%s.pickle' % cluster)
             cl = repo.get(cluster)
             if cl._patches:
                 print("Attributes changed: ")
-                for attr, val in cl._patches.items():
+                for attr, val in list(cl._patches.items()):
```
yarikoptic:
```suggestion
                for attr, val in cl._patches.items():
```

setup.py
```diff
@@ -17,10 +17,22 @@
...
+    in ['1', 'y', 'yes', 'true', 'on']):
+    force_install = True
+else:
+    force_install = False  # default
+
+
 import sys
 python_version = sys.version_info[:2]
-if python_version != (2, 7):
+if python_version != (2, 7) and not force_install:
```
yarikoptic: Why to require force_install for python3 - why not to strip this check entirely away?

elasticluster/providers/ec2_boto.py
```diff
@@ -52,6 +59,20 @@
...
+# over the former gives characters (i.e., `str` of length 1),
+# whereas iterating over the latter gives small integers.  I
+# could find no simple incantation that works on both, so
+# here's a conditional definition...
+if sys.version_info[0] == 2:
+    def byte_to_hex(byte):
+        return byte.encode('hex')
+else:
+    def byte_to_hex(byte):
+        return '{:02x}'.format(byte)
```
yarikoptic: in https://stackoverflow.com/questions/6624453/whats-the-correct-way-to-convert-bytes-to-a-hex-string-in-python-3 found `bytes.hex()`:
```
(Pdb) p bytes.hex
<method 'hex' of 'bytes' objects>
*(Pdb) p bytes.hex(pkey.get_fingerprint())
'4d745ca74dbad95bc7b4a730188967bc'
```
so if you had `bytes_to_hex`, then could be just `bytes_to_hex = bytes.hex` for py3.

elasticluster/subcommands.py
```diff
@@ -411,9 +417,9 @@ def execute(self):
             #       complicated for the user
             if (not grp in cluster.nodes or not cluster.nodes[grp]) \
                     and not template:
-                print "Elasticluster can not infer which template to use for "\
+                print("Elasticluster can not infer which template to use for "\
```
yarikoptic: minor: training `\` is no longer needed here since `()`

elasticluster/cluster.py
riccardomurri: > I don't think that list() is needed here (and there is a similar case below)

Yes, it is.  The loop modifies the `extra` dictionary, so we need to iterate on a *copy* of the keys (Python 2 semantics for `.keys()`); iterating over a *view* of the actual keys (Python 3 semantics for `.keys()`) raises a `RuntimeError`:

```python
>>> D = {'a':1, 'b':2, 'c':3}
>>> for key in D.keys():
...   if (D[key] % 2) == 0:
...     del D[key]
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
```

riccardomurri: > I would recommend to do first extra = extra.copy() right before the loop or instead of modifying extra in-place

While this is in general excellent advice, it's not needed in this case: `extra` is the "kwargs" part of the function signature, so changes to it won't be propagated back to the caller.  Exhibit::

```python
>>> D = {'a':1, 'b':2, 'c':3}
>>> def f(**extra):
...   del extra['b']
...   print(extra)
...
>>> f(**D)
{'a': 1, 'c': 3}
>>> print(D)
{'a': 1, 'b': 2, 'c': 3}
```

elasticluster/cluster.py
riccardomurri: Same reasoning as above. :-)

elasticluster/conf.py
riccardomurri: Good catch! ;-)

elasticluster/conf.py
riccardomurri: Yes.

elasticluster/migration_tools.py
riccardomurri: Yes. (But I should probably delete the entire file... I don't think there is anyone still around running such an old version of ElastiCluster that they still need this conversion script.)

setup.py
riccardomurri: You're right -- thanks for spotting it!  (It was a safeguard while I was working on the Py3 conversion, but by the time I was done, I had forgotten about it...)

elasticluster/providers/ec2_boto.py
riccardomurri: Good catch, many thanks for the suggestion!

elasticluster/cluster.py
yarikoptic: re `list()` - I was blind, although I have looked! indeed. sorry about the noise

re `copy()` - indeed!  I got a bit misguided by [us needing similar fix in datalad recently](https://github.com/datalad/datalad/commit/b09678336346805254b3f0319e1ac6e78994d4f6), but there we did deal with a dict of the kwargs as handled by wrapt... so indeed not the case generally/here.  Sorry for the noise again.

Having said both of those, this pattern used in two places (at least), so I wonder if in the long(er) run it is worth of having a helper method (`Struct._get_filtered_extras` or alike) which could possibly also `log.debug` (or `info`?) the fact that some extra fields are simply ignored?

riccardomurri:
With my current code, there are interoperability problems with Ansible
when running in Python 2.7.  The issue seems to be that `future`
installs a `newstr` type for literal strings, and this causes YAML
serialization to use this custom type which Ansible does not know how
to load:

```
ERROR! Syntax Error while loading YAML.
  could not determine a constructor for the tag 'tag:yaml.org,2002:python/object/new:future.types.newstr.newstr'

The error appears to have been in '/tmp/elasticluster.fAu_lD.d/extra_vars.yml': line 6, column 24, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  cloud:
    aws_access_key_id: !!python/object/new:future.types.newstr.newstr [!!python/unicode 'ABCDEFGHILJK...']
                       ^ here

```

riccardomurri:
@yarikoptic many thanks for your review and suggestions!  I have
merged them, will push to GitHub later (doing a force-push since I
rebased on top of current "master").

> main concern is that tests coverage seems to be quite "lean" 35%, so I guess there would be more daemons to tame

Yeah, the real problems will be in the actual interaction with Ansible
and with cloud providers, which is difficult to test (unless running
the actual thing).

riccardomurri:
It looks like *future*'s `newstr` feature is still deemed experimental and can be turned off by removing the lines `from builtins import str`; see: https://python-future.org/what_else.html#str

elasticluster/cluster.py
riccardomurri: > his pattern used in two places (at least), so I wonder if in the long(er) run it is worth of having a helper method

Good idea, but in the long(er) run I would actually like to get rid of the current loose constructors and tighten up classes `Cluster` and `Node` ... So, not sure if it's worth doing.  Anyway, not in scope for this PR which I'd like to keep focussed on the minimal modifications needed to get ElastiCluster running on Py3 :-)

riccardomurri:
Ansible interop problems udner Py2 seem to be solved in commit 03bf70d7

riccardomurri:
AWS EC2 still seems to have problems in Py2:

```
2019-04-02 17:11:31 monia gc3.elasticluster[13267] DEBUG Updating the ip addresses of `master001`.
2019-04-02 17:11:31 monia gc3.elasticluster[13267] ERROR Error: 'NoneType' object has no attribute 'get_ips'
Traceback (most recent call last):
  File "/home/rmurri/w/elasticluster/elasticluster/__main__.py", line 206, in main
    return self.params.func()
  File "/home/rmurri/w/elasticluster/elasticluster/subcommands.py", line 91, in __call__
    return self.execute()
  File "/home/rmurri/w/elasticluster/elasticluster/subcommands.py", line 759, in execute
    frontend.update_ips()
  File "/home/rmurri/w/elasticluster/elasticluster/cluster.py", line 1466, in update_ips
    self.ips = self._cloud_provider.get_ips(self.instance_id)
AttributeError: 'NoneType' object has no attribute 'get_ips'
Aborting because of errors: 'NoneType' object has no attribute 'get_ips'.
```

Not sure why... the YAML looks correct:

```
$ cat ~/.elasticluster/storage/ge16a.yaml
availability_zone: ''
extra:
    accelerator_count: 0
    allow_project_ssh_keys: true
    flavor: t2.micro
    image_id: ami-0ff760d16d9497662
    local_ssd_count: 0
    local_ssd_interface: SCSI
    login:
        image_sudo: true
        image_user: centos
        image_user_sudo: root
        image_userdata: ''
        user_key_name: rmurri
        user_key_private: ...
        user_key_public: ...
    master_nodes: '1'
    security_group: default
    worker_nodes: '1'
name: ge16a
nodes:
    master:
    -   cluster_name: ge16a
        extra:
            accelerator_count: 0
            local_ssd_count: 0
            local_ssd_interface: SCSI
            login: centos
            min_num: 1
        flavor: t2.micro
        image_id: ami-0ff760d16d9497662
        image_user: centos
        image_userdata: ''
        instance_id: i-02ea3286b5ba4161b
        ips:
        - 172.31.21.187
        - 34.243.227.82
        kind: master
        name: master001
        preferred_ip: 34.243.227.82
        security_group: default
...
```

elasticluster/providers/ec2_boto.py
```diff
@@ -69,8 +69,7 @@
     def byte_to_hex(byte):
         return byte.encode('hex')
 else:
-    def byte_to_hex(byte):
-        return '{:02x}'.format(byte)
+    byte_to_hex = bytes.hex
```
yarikoptic: oh sorry I was too cryptic in my comment by saying "if you had `bytes_to_hex`" (not just `byte_to_hex`), i.e. you could apply it to the full "string" of bytes but not to each individual integer which you get by traversing at the point of use of byte_to_hex ATM (used in a single spot iirc). and that is what causes travis to puke now
```
elasticluster/providers/ec2_boto.py:284: in start_instance
    self._check_keypair(key_name, public_key_path, private_key_path)
elasticluster/providers/ec2_boto.py:593: in _check_keypair
    for byte in pkey.get_fingerprint())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.0 = <bytes_iterator object at 0x7f332c006438>
    fingerprint = ':'.join(byte_to_hex(byte)
>                          for byte in pkey.get_fingerprint())
E   TypeError: descriptor 'hex' requires a 'bytes' object but received a 'int'
```

riccardomurri: Sorry, my bad for not checking even `bytes.hex()` signature...  Reverted the change, code should be correct on Py3 again now.

riccardomurri:
Well, the code now runs again on Py2 as well. I guess it's ready for merging into "master" but I'll leave it around for a few days, in case somebody volunteers to do a bit more testing.

riccardomurri:
Merged outside GitHub.

ElastiCluster has Python3 support, yay!

PR #625, opened 2019-04-01
ENH: Submit coverage report to codecov

yarikoptic:
Since the repository is public, this service is free and should be very easy to enable.  Someone with sufficient privileges should just login through github and enable codecov report for this repo. 
With [sourcegraph](https://github.com/codecov/sourcegraph-codecov/blob/master/README.md) browser extension it becomes then really handy to see what lines of diff in PR are covered by tests and which not.


riccardomurri:

PR #627, opened 2019-04-04
BF: pip install codecov; submit to codecov before docker operations

yarikoptic:
docker manipulation commands  could exit prematurely. We do want to submit to codecov even in the case
when it is a PR from outside contributor (thus would not be considered "trusted build")

This is a follow up to half-cooked #625 . Sorry I got back to it too late to check on it to only discover that I have forgotten to install it and have not mentioned "exit" statements before added by me line

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
 
 install:
   - "pip install --upgrade 'setuptools' 'pip>=8.1.2'"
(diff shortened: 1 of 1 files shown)


yarikoptic:
here we go! since no prior reports were submitted, the current one is just a baseline

riccardomurri:
Looks nice, thanks a lot!

PR #628, opened 2019-04-04
ENH(TST): smoke testing importing of the modules which otherwise not tested

yarikoptic:
spotted while reviewing #623 . although not that much of testing, but could be helpful.


riccardomurri:
Great idea, many thanks for taking the time to contribute it!

PR #632, opened 2019-04-11
Encode cluster and node names on Azure.

manuelesimi:
This fixes the following deployment error on Azure emerged after the migration to Python 3:

```
gc3.elasticluster[14] ERROR Could not start node `compute001`: Unicode-objects must be encoded before hashing -- <class 'TypeError'>
Traceback (most recent call last):
  File "/elasticluster-src/elasticluster/cluster.py", line 580, in _start_node
    node.start()
  File "/elasticluster-src/elasticluster/cluster.py", line 1318, in start
    **self.extra)
  File "/elasticluster-src/elasticluster/providers/azure_provider.py", line 292, in start_instance
    cluster_name, node_name)
  File "/elasticluster-src/elasticluster/providers/azure_provider.py", line 330, in _make_storage_account_name
    algo.update(cluster_name)
TypeError: Unicode-objects must be encoded before hashing
```

### DIFF
--- a/elasticluster/providers/azure_provider.py
+++ b/elasticluster/providers/azure_provider.py
@@ -320,8 +320,8 @@ class AzureCloudProvider(AbstractCloudProvider):
     @staticmethod
     def _make_storage_account_name(cluster_name, node_name):
         algo = hashlib.md5()
-        algo.update(cluster_name)
-        algo.update(node_name)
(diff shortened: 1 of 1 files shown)


manuelesimi:
I forgot to mention: 
I made the tests on Azure within a Docker container based on python:3.7.3-slim.

PR #633, opened 2019-04-11
Azure deployment: new template, managed disks and disk boot size

manuelesimi:
I extended the Azure deployment to use managed disks on the nodes and added the possibility to set the boot disk size with a new parameter (boot_disk_size) in the cluster configuration file.

The API version has been upgrade to 2018-10-01.

My tests have been made on the region eastus2. Unfortunately, not all the API versions are available in all the regions. But this was true also for the previous deployment template based on version 2015-06-15.


riccardomurri:
> Unfortunately, not all the API versions are available in all the regions.

Is there a document or web page that shows what API versions are available for each zone?

elasticluster/share/etc/azure_vm_template.json
```diff
@@ -88,23 +95,47 @@
     "nicName": "[concat(parameters('vmName'), '-nic')]",
     "publicIPAddressName": "[concat(parameters('vmName'), '-public-ip')]",
     "publicIPAddressType": "Dynamic",
-    "storageAccountType": "Standard_LRS",
```
riccardomurri: This, together with the new line 270, changes the storage account type to `StandardSSD_LRS`. I guess the price tag is different for the two types.  Would it make sense to have this as a configurable option?

elasticluster/share/etc/azure_vm_template.json
```diff
@@ -88,23 +95,47 @@
...
-        "accountType": "[variables('storageAccountType')]"
+        "networkAcls": {
+          "bypass": "AzureServices",
+          "virtualNetworkRules": [],
+          "ipRules": [],
+          "defaultAction": "Allow"
+        },
+        "supportsHttpsTrafficOnly": false,
+        "encryption": {
+          "services": {
```
riccardomurri: Is this relevant for VM disks? or does it only apply to object storage?

elasticluster/share/etc/azure_vm_template.json
```diff
@@ -227,15 +259,16 @@
             "publisher": "[parameters('imagePublisher')]",
             "offer": "[parameters('imageOffer')]",
             "sku": "[parameters('imageSku')]",
-            "version": "latest"
+            "version": "[parameters('imageVersion')]"
```
riccardomurri: Nice catch, thanks!!

riccardomurri:
LGTM, thanks!  If you can answer the few questions raised in line-by-line comments, I can add a couple of notes to the docs and merge.

elasticluster/share/etc/azure_vm_template.json
manuelesimi: That's correct. There are four options for Azure managed disks:
- Ultra SSD Managed Disks
- Premium SSD Managed Disks 
- Standard SSD Managed Disks
- Standard HDD Managed Disks

Their price varies according to the region, subscription type and redundancy:
https://azure.microsoft.com/en-us/pricing/details/managed-disks/

A parameter for the type would be appropriate.

elasticluster/share/etc/azure_vm_template.json
```diff
@@ -88,23 +95,47 @@
     "nicName": "[concat(parameters('vmName'), '-nic')]",
     "publicIPAddressName": "[concat(parameters('vmName'), '-public-ip')]",
     "publicIPAddressType": "Dynamic",
-    "storageAccountType": "Standard_LRS",
     "networkSecurityGroupName": "[concat(resourceGroup().name, '-secgroup')]",
     "sshKeyPath": "[concat('/home/',parameters('adminUsername'),'/.ssh/authorized_keys')]",
     "vnetID": "[resourceId('Microsoft.Network/virtualNetworks', parameters('virtualNetworkName'))]",
     "subnetRef": "[concat(variables('vnetID'),'/subnets/',parameters('subnetName'))]",
-    "apiVersion": "2015-06-15"
+    "apiVersion": "2018-10-01"
```
riccardomurri: According to https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-supported-services API versions vary *per zone* (as you point out) and *per service*... This makes it kind of unwieldy to hard-code a number here, as that may seemingly randomly fail.

Would it be possible to have something like `latest` here instead of a specific API version number?

manuelesimi:
> > Unfortunately, not all the API versions are available in all the regions.
> 
> Is there a document or web page that shows what API versions are available for each zone?

My understanding is that supported API versions are sort of dynamic and they are even different for each resource type.

Let’s consider the deployment template I added with this PR. As you notice, the api version used for the storageAccounts (line 108) resource is not the same as the general one for the template (line 102). If I change line 108 to:
"apiVersion": "[variables('apiVersion')]",

I get the following error:

```
msrestazure.azure_exceptions.CloudError: Azure Error: DeploymentFailed
Message: At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details.
Exception Details:
        Error Code: BadRequest
        Message: {'error': {'code': 'NoRegisteredProviderFound', 'message': "No registered resource provider found for location 'eastus2' and API version '2018-10-01' for type 'storageAccounts'. The supported api-versions are '2019-04-01, 2018-11-01, 2018-07-01, 2018-03-01-preview, 2018-02-01, 2017-10-01, 2017-06-01, 2016-12-01, 2016-05-01, 2016-01-01, 2015-06-15, 2015-05-01-preview'. The supported locations are 'eastus, eastus2, westus, westeurope, eastasia, southeastasia, japaneast, japanwest, northcentralus, southcentralus, centralus, northeurope, brazilsouth, australiaeast, australiasoutheast, southindia, centralindia, westindia, canadaeast, canadacentral, westus2, westcentralus, uksouth, ukwest, koreacentral, koreasouth, francecentral, southafricanorth'."}}
```

This is because my region (eastus2) does not support the API 2018-10-01 for the storage resource, while it does for all the other resource types in the template. Hence, I had to use a different API version (and template section) for the storage.

Since the supported API versions are dynamic, one needs to discover which API versions are available in the target region. 

There are two ways to do that:

1. You try, get the error above and then change the template accordingly
2. You explore the available resource/api as described here: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-register-provider-errors

The **biggest drawback** here is that the format of our deployment template would change, depending on the API versions used. To be more explicit, the format of the JSON for a resource type is different from one API version to the other. 

I see two possible workarounds here:
1. add an optional parameter for EC that can be used to pass a custom VM template
2. document where is the VM template and how to change it (https://docs.microsoft.com/en-us/azure/templates/)

elasticluster/share/etc/azure_vm_template.json
manuelesimi: I wish it would be possible, but it is not. As I wrote in the other comment, the format of the JSON template is different from one API version to another). So there is no way to have a general template for all the regions.
In fact, your previous WM template was not working in my region either.

elasticluster/share/etc/azure_vm_template.json
manuelesimi: If you mean _supportsHttpsTrafficOnly_, it is a parameter of the storage account resource. It allows https traffic only to storage service if sets to true. It is requested by the new API 2018-07-01 used for this resource.
I thought a default to false would be more common than true. Feel free to change it if you wish so.

elasticluster/share/etc/azure_vm_template.json
manuelesimi: :)

elasticluster/share/etc/azure_vm_template.json
riccardomurri: No, sorry -- I meant the "encryption" settings.

manuelesimi: It applies to the storage account. Again, I thought that enabling encryption for those services would be more common (or at least it doesn't hurt if you do not use them at all).
Here's the reference doc:
https://docs.microsoft.com/en-us/azure/templates/microsoft.storage/2018-07-01/storageaccounts

elasticluster/share/etc/azure_vm_template.json
riccardomurri: Valid values for this parameter are the ones listed in the following page, correct?
https://docs.microsoft.com/it-it/rest/api/compute/disks/createorupdate#disksku

elasticluster/share/etc/azure_vm_template.json
riccardomurri: Makes good sense :-)

elasticluster/share/etc/azure_vm_template.json
manuelesimi: Correct.

riccardomurri:
Re: use of API version: It seems that MS is aware of the issue, and they recommend the use of "API profiles" to create ARM templates that work across zones?  At least this is my reading of https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates-cloud-consistency#track-versions-using-api-profiles

manuelesimi:
It's unclear to me how profiles could work across api versions when the JSON format is different for the same resource but with a different API.

I could give it a try in the next days. For instance, a good test will be to see if the storage account API that is missing in my region can be resolved with using the API profile. 

Also, I'm afraid  that they just moved up the issue at profile level. What would happen if the profile API version in the template is not available in the target region?

riccardomurri:
My understanding of API profiles is that they're guaranteed to be available across all zones and services...

manuelesimi:
You are right. It seems that they have 3 API profile versions across all the regions:

https://github.com/Azure/azure-rest-api-specs/tree/master/profile

More doc here:
https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-version-profiles

manuelesimi:
I started to play with API profiles and I report here my early feedback.

As I suspected, the API profiles do not resolve the issues that MS introduced by supporting different APIs in different regions for different resource types (!).

First, not all the deployment schema templates are available in all the regions. I started with the latest available, and I got the following error:

```
msrestazure.azure_exceptions.CloudError: Azure Error: InvalidTemplate

Message: Deployment template validation failed: 'Template schema 'https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#' is not supported. Supported versions are '2014-04-01-preview,2015-01-01,2018-05-01'. Please see https://aka.ms/arm-template for usage details.'.
```

So, the 2019-04-01 is not supported in my region. I downgraded to use the 2018-05-01 schema.

Now, each template deployment schema has an enumeration of supported API profiles (typically 3). I picked the _2018-03-01-hybrid_ API profile, the most recent of the 2018-05-01 schema, and made my tests with this version.

And here I got what I was afraid to get, i.e. errors related to the wrong format of my deployment template:

```
msrestazure.azure_exceptions.CloudError: Azure Error: DeploymentFailed

Message: At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details.
Exception Details:
        Error Code: BadRequest
        Message: {'error': {'code': 'InvalidRequestParameters', 'message': 'Request parameters are invalid: encryption.services.file.'}}
```

Or:

```
Message: At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details.
Exception Details:
        Error Code: BadRequest
        Message: {'error': {'code': 'InvalidParameter', 'message': "Storage account type 'StandardSSD_LRS' is supported by Microsoft.Compute API version 2018-04-01 and above", 'target': 'osDisk.managedDisk.storageAccountType'}}
```

These errors mean that the resource APIs resolved with the profile are not compatible with my resource definitions. **And unfortunately it makes totally sense**. 

The way I understand API profiles, they are the equivalent of a Bill of Materials in Maven. In Maven, you don’t declare the version of a dependency in your artifact’s POM and you let the BOM resolve it. BOMs are created to make sure that the collected artifacts work well together. I suppose Microsoft created the API profiles as collection of API versions that work in all (or most) of the regions. But it’s my guess. 

But, to continue the mirroring with BOMs, if you advance the BOM version, your source code could stop compiling because it is not compatible with the new dependencies resolved by the BOM. The same happens with API profiles. If we have a resource A compatible with a schema version X, then A is not compatible with all the API profiles, but only with the ones that support version X.

At this time, I didn’t find yet a deployment template schema that supports an API version >= 2018-10-01, the ones we need for managed disks.

Is this a dead end for the API profile option? No, at least not yet. I’ll investigate more tomorrow and give you another update.

riccardomurri:
Wow, such a thorough investigation!  Many many thanks for putting time and effort into this!

manuelesimi:
I think I found a good compromise to use one of the most recent API profiles and schema templates. More specifically:

1) The network and vm templates declare and are compatible with the _2018-05-01_ schema template (the one used before was from 2014).

2) Both templates now use the API profile version _2018-03-01-hybrid_.

3) No API version is declared in any resource declaration.

To make thinks working with the above configuration I had to tweak some resources:

- the encryption for Files has been removed from the Storage account resource. It was introduced starting from the API version 2016-12-01, while our API profile uses the API version 2016-01-01. I think this is file within the scope of this PR.

- Storage accounts  of type _StandardSSD_LRS_ are supported by Microsoft.Compute API version 2018-04-01 and above. Again, it has been downgraded to _Standard_LRS_. This rules out the possibility to have the account type as parameter. It is bound to the API and cannot be arbitrarily changed by the user.

Other than that, I was able to successfully create 4 clusters with different configurations using the new templates.

This is the declaration of the API profile used:
https://github.com/Azure/azure-rest-api-specs/blob/master/profile/2018-03-01-hybrid.json
It would be good to refer it in the EC documentation to have a complete list of the supported API versions for each resource type. Again, I assume that API profiles are created with list of APIs supported in all the regions, but I can't prove that.

If this PR is accepted, I suggest to revise the documentation to explain how to customize the VM and Net templates by largely referring the official Azure webpages. 
Also, we should make it clear that EC can create clusters wherever the API profile _2018-03-01-hybrid_ is supported (I believe it’s way better than explains it resource by resource….).
I can help expand the doc, if you wish so.

manuelesimi:
I performed some tests in the following regions:
-  `westus`
-  `southcentralus`
-  `westeurope`
-  `eastus2`

The resource APIs declared within the API profile version 2018-03-01-hybrid were accepted in all these regions. Therefore, I was able to create and configure my clusters with the new templates in all of them.

riccardomurri:
Ok, so my proposal for this PR would be the following:

1. merge the PR as is now,
2. add an option to set the disk storage type (I guess not just `Standard_LRS` is supported in the "hybrid" profile?)
3. add an option to override the ARM template with a custom one provided by the user.

I guess 3. would only be useful to very technical users, but still if one knows how to unlock features (e.g., encryption) then this could be done.  A possible downside is that ameliorations to the ARM template shipped with ElastiCluster become a matter of local configuration and will likely not be contributed back upstream...

Comments?

riccardomurri:
Regarding:

> I suggest to revise the documentation to explain how to customize the VM and Net templates by largely referring the official Azure webpages. Also, we should make it clear that EC can create clusters wherever the API profile 2018-03-01-hybrid is supported 

Can you please add the required edits to the documentation to this PR?  It's rather clear that you have a much better picture than me about what it takes to build a cluster on Azure...

manuelesimi:
> Ok, so my proposal for this PR would be the following:
> 
> 1. merge the PR as is now,
> 2. add an option to set the disk storage type (I guess not just `Standard_LRS` is supported in the "hybrid" profile?)

May be it's worthy to have the option exposed. Possible values for the 2017-03-30 API are:
- Standard_LRS
- Premium_LRS

From: https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/2017-03-30/virtualmachines

The availability of the premium disks depends on the subscription and the selected VM (not all the VMs support the premium storage).

If one needs more performing Solid State Drives (SSD), the ARM template has to be changed to use a more recent API for the Compute resource (not supported by the 2018-03-01-hybrid API). And, of course, we can document that too.

> 3. add an option to override the ARM template with a custom one provided by the user.

I agree.

> 
> I guess 3. would only be useful to very technical users, but still if one knows how to unlock features (e.g., encryption) then this could be done. A possible downside is that ameliorations to the ARM template shipped with ElastiCluster become a matter of local configuration and will likely not be contributed back upstream...

In my opinion, if we fully embrace API profiles as a general mean to provide the most common computational clusters, the only improvement worthy to integrate would be an upgrade to a more recent API profile released by Microsoft. I see any other customization as a way to serve specific requirements of projects that EC cannot reasonably cover. If we start supporting corner cases or write our own API profiles, it would be a project in its own and, in any case, we would never be able to cover all the provides in all the regions.

> 
> Comments?

Sounds a good plan to me.

manuelesimi:
> Regarding:
> 
> > I suggest to revise the documentation to explain how to customize the VM and Net templates by largely referring the official Azure webpages. Also, we should make it clear that EC can create clusters wherever the API profile 2018-03-01-hybrid is supported
> 
> Can you please add the required edits to the documentation to this PR? It's rather clear that you have a much better picture than me about what it takes to build a cluster on Azure...

Sure, I can do that. My understanding is that these edits should only cover the changes in this PR. The new options for the custom ARM template and the storage account type should be documented later (when the options are added to EC). Is that correct?

riccardomurri:
>  if we fully embrace API profiles as a general mean to provide the most common computational clusters, the only improvement worthy to integrate would be an upgrade to a more recent API profile released by Microsoft.

Absolutely agree!

> I see any other customization as a way to serve specific requirements of projects that EC cannot reasonably cover. If we start supporting corner cases or write our own API profiles, it would be a project in its own and, in any case, we would never be able to cover all the provides in all the regions.

Does this mean you think that adding an option for customized VM deployment templates is a good idea, or is it not worth pursuing?

manuelesimi:
Sorry for being not clear in the comment. I do think it's a great idea to add the option for custom ARM templates. I would definitely go for it.

My observation was related to your concern of loosing improvements by opening up the ARM template to users.  Since we cannot cover all the possible requirements of individual projects,  I think that supporting the official API profiles from Microsoft is the best we can do in EC. In this direction,  adding support for future profiles is the only improvement worthy to be integrated.

riccardomurri:
Good :-) I have implemented the two changes on top of this PR, in my branch https://github.com/riccardomurri/elasticluster/tree/pr/%23633

I'll make a pull request to this PR, so you can merge those changes and test -- let me know if everything still works (it does in my tests, but they are quite limited), then I'll merge.

riccardomurri:
Hmmm... I've done the PR, but now all commits from recent master show up there...  Maybe it's better if we both rebase against current "master"?

manuelesimi:
Yep, I got all the commits from master with the PR (I read your latest comment too late..). Let me try to clean up the branch by rebasing.

manuelesimi:
Not much luck with the rebasing or the merge from your master branch. What if I revert my Azure_disks branch to the commit before your PR and then apply your changes for the new options?

riccardomurri:
Iǘe done the rebase in my branch pr/#633 -- can you pls try that out?  If it works, I can merge that instead.

manuelesimi:
> Iǘe done the rebase in my branch pr/#633 -- can you pls try that out? If it works, I can merge that instead.

Great. I will pull your branch and test it.

manuelesimi:
I cloned your repo and used the 633 branch to make some tests. All good. Green light from my side for the merge. 

I also noticed you already edited the documentation. That's great. I just have a minor comment.

The following is probably a better (and more human readable) link for the 2018-03-01-hybrid profile:

https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-profiles-azure-resource-manager-versions#overview-of-2018--03-01-hybrid

Thanks for your support all along!

riccardomurri:
Merged into "master" now.

Many many thanks for taking care of updating the Azure support and the fruitful discussion!

Issue #635, opened 2019-04-19
no integrity check


brokenquark:
Greetings,

I am a security researcher, who is looking for security smells in Ansible scripts. I found instances where no integrity check is used after downloading an iso or package binaries. According to the Common Weakness Enumeration organization this is a security weakness (Missing Support for Integrity Check: https://cwe.mitre.org/data/definitions/353.html). 

Any feedback is appreciated.

source: https://github.com/gc3-uzh-ch/elasticluster/blob/master/elasticluster/share/playbooks/roles/jupyter/tasks/pyspark.yml

riccardomurri:
Thanks for taking the time to audit ElastiCluster sources for security
issues!

In this specific case, though, I'm uncertain about what the issue is: you
mention a missing integrity check after download, but there is no download
action in the Ansible file you quoted...

brokenquark:
Hi, 
thanks for your quick response. In line 12 and 26, checksum is not checked. 

We are trying to build an automated tool that can catch these types of cases when no integration check is performed. In your case, local python binary is used without checking. 

Our target is to only detect any coding practice that can lead to exploit or vulnerability. However, in this particular case, it won't lead to any exploit in my humble opinion. 

However, according to CWE, any binary should be gone through an integrity check whenever necessary or unnecessary. We are just trying to build our tools according to that suggestions. 

Thanks again :)

riccardomurri:
> In line 12 and 26, checksum is not checked.

I think there's a misunderstanding about what the actions in that playbook are doing. The `get_checksum: no` lines belong to a `stat:` action, which gathers information about a file on the filesystem of the machine being configured.  In this case, the playbook is only checking whether the files *exist* (in order to decide whether to install the relevant Jupyter kernel description) so there's no need to compute MD5 checksum: we just need to probe if the file is there.

> In your case, local python binary is used without checking.

The local Python binary is supposed to be part of the base OS.  We have to trust something :-)

PR #640, opened 2019-05-09
Update pyspark.yml

akondasif:
Changed `get_checksum` to yes to support integrity check. Without integrity check downloaded content cannot be validated.


riccardomurri:
I'm not sure what's the purpose of this?  That Ansible stanza only checks for the presence of `pyspark`, in order to determine wheter to install the PySpark Jupyter kernel (line 59).  What would computing the checksum add to the process?

See a more detailed discussion to a similar (same?) issue in #635 and particularly https://github.com/gc3-uzh-ch/elasticluster/issues/635#issuecomment-484995671

PR #641, opened 2019-05-27
Update mgr.yml to avoid error with ansible 2.8

julienperret:
Ansible 2.8.0 complains with the syntax of mgr.yml
"ERROR! 'become' is not a valid attribute for a TaskInclude"
We are now using ansible 2.8 syntax (is it actually 2.7 syntax?).


riccardomurri:
Many thanks for taking the time to fix this and make a PR!

It looks like the `apply: ...` syntax is supported from Ansible 2.7 onwards, but ElastiCluster currently only requires >=2.5 -- can you please also update the requirement in `setup.py`?  I will merge as soon as that is done.

riccardomurri:
Thanks a lot for contributing and for the quick reaction!

julienperret:
You are very welcome!
Thank you for the awesome software!

PR #643, opened 2019-06-06
Update glusterfs.yml to take into account glusterfs variables stripes and replicas

julienperret:
After spending some time trying to create Replicated Volumes, I found out that the variables replicas and stripes were swapped.
I also had to use variable:
compute_var_glusterfs_replicas=4

whereas the documentation (https://elasticluster.readthedocs.io/en/latest/playbooks.html?highlight=gluster_replicas) stated:
compute_var_gluster_replicas=4

I just wanted to make sure I was right before also making a PR on docs/playbooks.rst and examples/glusterfs.conf

### DIFF
--- a/docs/playbooks.rst
+++ b/docs/playbooks.rst
@@ -894,11 +894,11 @@ and any ``glusterfs_client`` to mount this filesystem over directory
 ``/glusterfs``.
 
 To manage the GlusterFS filesystem you need to connect to a
-``gluster_server`` node.
+``glusterfs_server`` node.
 
 Several versions of GlusterFS are concurrently maintained and packaged upstream;
 the actual version of GlusterFS installed by ElastiCluster can be set using global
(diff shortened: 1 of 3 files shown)


riccardomurri:
I'm terribly sorry!  This is the kind of trivial error which can just cost you hours of time to debug... :-(

riccardomurri:
Many thanks for this patch!  If you could update the docs, I will accept it right away.  (The docs should reflect the names as actually used in the playbooks; doing the opposite would break the configs of those already using it.)

julienperret:
I also updated variable glusterfs_redundancy even though I couldn't find where, in the playbook, the variable was used. I suspect it might not be but it also might be that I'm out of my depth.

riccardomurri:
Many many thanks for solving this and taking care of examples and docs!  I wish had more contributors this thotough :-)

julienperret:
You are very welcome :-)

PR #644, opened 2019-06-06
Implement availability zone support for OpenStack.

riccardomurri:
This is draft code for implementing AZ support in the OpenStack cloud backend.

To use it, add the `availability_zone` configuration key setting to the OpenStack cloud stanza:

```
[cloud/openstack]
# ... other settings stay the same ...
availability_zone = az1
```

Note that ElastiCluster will *not* verify if the AZ name is correct; any mis-spelling or wrong names will just result in nodes not starting with an error.

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -495,6 +495,11 @@ Valid configuration keys for ``openstack``
   is set, this option is ignored and the value of the environment
   variable is used instead.
 
+``availability_zone``
+  If this is given, all cluster nodes will be started in this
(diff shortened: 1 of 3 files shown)


riccardomurri:
I cannot test this code myself, as I do not have access to an OpenStack infrastructure with AZs -- any reports of testing, what works, what doesn't is very welcome!

riccardomurri:
I have built a Docker image for this, tagged `pr644`:

```
./elasticluster.sh --release pr644
```

PR #669, opened 2019-11-26
disable package upgrade on redhat when upgrade_packages=no

pescobar:
do the same as for [Debian systems](https://github.com/elasticluster/elasticluster/blob/master/elasticluster/share/playbooks/roles/common/tasks/init-Debian.yml#L67-L71)


riccardomurri:
Merged, many thanks for taking the time to patch and contribute!

Issue #248, opened 2016-04-22
Cannot add a compute to a frontend-only cluster


smaffiol:
* with a frontend only cluster:
$ elasticluster resize -a 1:compute

Error:
Adding 1 compute node(s) to the cluster
Elasticluster can not infer which template to use for the new node(s). Please provide the template with the `-t` or `--template` option

PR #673, opened 2020-01-23
allow to start 0 instances

### LINKED ISSUE #672: Starting 0 instances
I want to create a slurm cluster with different node types (e.g., [here](https://github.com/elasticluster/elasticluster/blob/master/examples/slurm-with-different-node-types.conf)). 

However, on `start` I only want to boot smallworkers and later on if needed `resize` to add largeworkers.
So the config file would have
```
largeworker_nodes=0
```

At the moment this is not possible and will give on `start`
```
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/elasticluster/cluster.py", line 695, in _check_cluster_size
    available = len(self.nodes[kind])
KeyError: 'largeworker'
```

How about changing [L695](https://github.com/elasticluster/elasticluster/blob/dcb9ec3587d5fa4921306fa034bf65dc57cf94af/elasticluster/cluster.py#L695) to 
`available = len(self.nodes.get(kind, 0))`

to allow for this behaviour?

fliem:
closes #672

Hi @riccardomurri,

This PR enables defining different worker types in the config file and setting `{XXXworker}_nodes=0`, so that they are not created on `start` , but can be added later on via `resize`.

I ran some tests on openstack configuring a slurm cluster -- works as expected.
One can start 0 instances, add, and remove all instances of one class.

Note that resizing from 0 requires a template (#248)

### DIFF
--- a/elasticluster/cluster.py
+++ b/elasticluster/cluster.py
@@ -692,7 +692,7 @@ class Cluster(Struct):
         # finding all node groups with an unsatisfied amount of nodes
         unsatisfied = 0
         for kind, required in min_nodes.items():
-            available = len(self.nodes[kind])
+            available = len(self.nodes.get(kind, []))
             if available < required:
                 log.error(
                     "Not enough nodes of kind `%s`:"


riccardomurri:

(and sorry that I didn't look into this in the promised timeframe...)

fliem:
No worries. Thank you and thanks for developing this tool.

PR #696, opened 2021-03-15
Add support for OpenStack application credentials

pansapiens:
I've tested this and it's working for me.

Supports `application_credential_id` and `application_credential_secret` in the elasticluster config, or via the OpenStack environment variables `OS_APPLICATION_CREDENTIAL_ID` and `OS_APPLICATION_CREDENTIAL_SECRET`.

### DIFF
--- a/elasticluster/conf.py
+++ b/elasticluster/conf.py
@@ -249,6 +249,8 @@ CLOUD_PROVIDER_SCHEMAS = {
         Optional("cacert"): existing_file,
         Optional("username"): nonempty_str,
         Optional("password"): nonempty_str,
+        Optional("application_credential_id"): nonempty_str,
+        Optional("application_credential_secret"): nonempty_str,
         Optional("user_domain_name"): nonempty_str,
         Optional("project_domain_name"): nonempty_str,
         Optional("project_name"): nonempty_str,
--- a/elasticluster/providers/openstack.py
+++ b/elasticluster/providers/openstack.py
@@ -117,6 +117,8 @@ class OpenStackCloudProvider(AbstractCloudProvider):
 
     :param str username: username of the keystone user
     :param str password: password of the keystone user
+    :param str application_credential_id: application credential id
+    :param str application_credential_secret: application credential secret token
     :param str project_name: name of the project to use
     :param str user_domain_name: name of the user domain to use
     :param str project_domain_name: name of the project domain to use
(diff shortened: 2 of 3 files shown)


elasticluster/providers/openstack.py
```diff
@@ -381,23 +394,42 @@ def __init_keystone_session_v3(self, check=False):
...
+                username=self._os_username,
+                password=self._os_password,
+                user_domain_name=self._os_user_domain_name,
+                project_domain_name=self._os_project_domain_name,
+                project_name=self._os_tenant_name,
+            )
+        else:
+            log.error("Must specify an OpenStack password or application_credential_id + "
+                      "application_credential_secret (likely configuration error?)")
+            raise ConfigurationError("No password or application_credential_id + application_credential_secret configured.")
```
riccardomurri, line 431:
```suggestion
            raise ConfigurationError("No password nor application credentials ID+Secret configured.")
```

elasticluster/providers/openstack.py
```diff
@@ -381,23 +394,42 @@ def __init_keystone_session_v3(self, check=False):
...
-            project_name=self._os_tenant_name,
-        )
+        if self._os_password and self._os_application_credential_id and self._os_application_credential_secret:
+            log.warning("Both password and application_credential_id/application_credential_secret are set. "
+                        "Ignoring password, using application credentials")
+
+        if self._os_application_credential_id and self._os_application_credential_secret:
+            auth = keystone_v3.ApplicationCredential(
+                auth_url=self._os_auth_url,
+                username=self._os_username,
```
riccardomurri, line 414: Isn't `username` redundant here?

elasticluster/conf.py
```diff
@@ -249,6 +249,8 @@ def make_config_parser():
         Optional("cacert"): existing_file,
         Optional("username"): nonempty_str,
         Optional("password"): nonempty_str,
+        Optional("application_credential_id"): nonempty_str,
```
riccardomurri, line 252: I understand that the `application_credential_id` is a long hex string.  If that is true, perhaps we can be stricter here and check that is all hexdigits and not just nonempty?

riccardomurri (approved):
LGTM, save for a few minor points.  Let me know if you don't have time to fix them; I'll merge as-is as it looks good anyway.

elasticluster/providers/openstack.py
pansapiens: Happy to make your suggested change - my intention here is to provide a more specific greppable string (application_credential_id and application_credential_secret) in the error message so that it's easier to search the docs/web for the missing config option. The log.error provides this, so not as important in the Python exception message, and I guess in the context of calling this programmatically that detail in the exception message makes less sense.

mberacochea:
This works for me, thank you @pansapiens and @riccardomurri :1st_place_medal:  . I'm running this with Openstack Ussuri (http://docs.embassy.ebi.ac.uk/userguide/Embassyv4.html#about-version-4), with a slurm cluster using Ubuntu Cloud 18.04.

elasticluster/providers/openstack.py
pansapiens: It is redundant. I've removed it and tested, works without it.

elasticluster/conf.py
pansapiens: I had a quick look to see if I could find a definitive spec from OpenStack as to whether this would _always_ be a hex string - all I found was that it should be globally unique to that cloud. I think it's safer to leave it for the moment since I can't be sure it isn't something that could have a different format between OpenStack deployments or releases.

pansapiens:
Sorry this took so long. I've updated the branch with minor changes as discussed.

PR #704, opened 2021-07-02
Load the key based on key type

onstring:
This is to fix the issue that paramiko does not raise exceptions when using the wrong key type


jeffctr:
shouldn't that fix be done in `paramiko` instead of here?

PR #706, opened 2021-07-19
Improve GPU support - add consumer card and bug fixes for centos 7.9

kandolfp:
Hey, I had a couple of issues to get GPUs working on Openstack with a Centos 7.9 image (7.8 as well) and with gpus not recognizing _consumer_ cards. The changes are documented within the commit messages.

### DIFF
--- a/elasticluster/share/playbooks/library/gpus
+++ b/elasticluster/share/playbooks/library/gpus
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Test run this module with::
 #
--- a/elasticluster/share/playbooks/library/gpus
+++ b/elasticluster/share/playbooks/library/gpus
@@ -181,7 +181,7 @@ def get_pci_gpu_info():
         if len(parts) < 3:
             # FIXME: log unexpected input!
             continue
-        if parts[1] == '0302' and parts[2] == '10de':
+        if (parts[1] == '0302' or parts[1] == '0300') and parts[2] == '10de':
             slot = _format_pci_slot(parts[0])
             vendor_name, model_name, model_name_long = get_pci_device_name(parts)
             result.append({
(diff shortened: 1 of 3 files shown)


elasticluster/share/playbooks/library/gpus
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
jeffctr, line 1: I think this code is written in pytohn2.7 and now you are forcing it to run with python3. I think you should upgrade the code to python3. There are not that many change to get it with python3 code style.

kandolfp: The only changes necessary from python2 to 3 would be to delete the python2 version of the import in line 63-68. The rest conforms to python3 code style (as far as I and 2to3 can find). Not sure if this needs to be deleted?
```
- try:
-     # Python 2
-     from urllib import urlopen
- except ImportError:
-     # Python 3+
+from urllib.request import urlopen
```

kandolfp: While searching for alternatives to this change I came across this description. 
https://britishgeologicalsurvey.github.io/devops/centos7-python2-end-of-life/

The problem I encountered was the same. Alternatively the code could be run explicitly with python3 and not setting it in the file. Not sure what is the better solution here.

jeffctr: Totally agreed. From the EC doc is mentioned that it keep working with python2. However, I think that it should change. I had to do some changes and I add this variables to my config file `global_var_ansible_python_interpreter=/usr/bin/python3`, and I had to install python3.

jeffctr: On the other hand I noticed that there are PRs hanging around for around a year. No idea what is going on with this Repository.

